PR #25, opened 2026-04-13
feat(cli): add --log-level flag for structured log verbosity control

cblecken-godaddy:
Adds a --log-level option (error, info, debug) as an alternative to --verbose/--debug/--info flags, with conflict detection and warnings.

Q: Whether we should add trace and warning via log_level command line option?

Alternative is to eliminate log_level from command line and the docs.


wcole1-godaddy:
What is the overall goal with this? Since this is OSS, we also need to be careful about our logs and anything that we might be outputting with them.

PR #28, opened 2026-04-23
fix(DEVX-68): reject localhost/private URLs in application init

asarkar2-godaddy:
Ticket: https://godaddy-corp.atlassian.net/browse/DEVX-68

Added `publicHttpUrl` — an arktype narrow backed by a pure `isPublicRoutableUrl` predicate. Strict subset of the old URL keyword: still requires a syntactically-valid URL, and additionally requires `http` / `https` scheme and a host outside the IETF/IANA non-routable classes:
- `localhost`, `*.localhost`, `*.local` (RFC 6761 / 6762)
- IPv4 loopback `127.0.0.0/8` and unspecified `0.0.0.0` (RFC 1122)
- IPv4 link-local `169.254.0.0/16` (RFC 3927)
- RFC 1918 private ranges `10/8`, `172.16/12`, `192.168/16`
- IPv6 loopback `::1`, unspecified `::`, link-local `fe80::/10` (RFC 4291)

Swapped arktype's built-in URL keyword for `publicHttpUrl` at both `application init` input schemas. Validation already ran first in the init effect, so bad URLs now short-circuit before any auth or HTTP call. Auth, HTTP, environment routing, and WAF rules are untouched.

The error message now includes the arktype summary — the failing field, the rejected value, and the reason — in place of the previous generic "Invalid application configuration".

Updated the documentation with a note on the updated URL constraints for `application init`

### DIFF
--- a/.changeset/devx-68-reject-private-urls.md
+++ b/.changeset/devx-68-reject-private-urls.md
@@ -1,0 +1,5 @@
+---
+"@godaddy/cli": patch
+---
+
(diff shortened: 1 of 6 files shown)


cblecken-godaddy (approved):
Looks good

wcole1-godaddy (approved):

PR #47, opened 2026-05-25
feat: allow application creation without required authorization scopes

dmisic-godaddy:
Made application creation without authorization scopes.


dmisic-godaddy:
Hey @wcole1-godaddy, please take a look, this is what we discussed about having apps without scopes, hence the change. I removed gates for scopes on app creation.

PR #64, opened 2026-06-15
feat: add gddy payments add command

sswaminathan-godaddy:
## Summary

- Adds `gddy payments add` — opens the system default browser to the GoDaddy payment methods management page
- URL is environment-aware: `prod` → `account.godaddy.com`, others → `account.{env}-godaddy.com`
- No CLI auth required; user authenticates in the browser if needed
- JSON output includes an `info` field confirming the browser was opened and noting that only credit card or Good-as-Gold can be used for domain purchases
- Extended `--help` text surfaces the same constraint for discoverability

## Test plan

- [ ] `cargo test payments::` — 4 unit tests covering URL derivation for prod/ote/dev/test
- [ ] `cargo clippy -- -D warnings` — clean
- [ ] `cargo run -- payments add --help` — shows long description with CC/GAG note
- [ ] `cargo run -- payments add` — opens browser to prod payment methods page
- [ ] `cargo run -- payments add --env ote` — opens browser to ote payment methods page

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1105,6 +1105,7 @@ dependencies = [
  "domains-client",
  "fancy-regex",
  "httpmock",
+ "open",
  "regex",
  "reqwest",
  "serde",
--- a/rust/src/payments/mod.rs
+++ b/rust/src/payments/mod.rs
@@ -1,0 +1,82 @@
+//! `gddy payments` — payment method management.
+
+use cli_engine::{
+    CliCoreError, CommandResult, CommandSpec, GroupSpec, Module, RuntimeCommandSpec,
+    RuntimeGroupSpec, Tier,
+};
+use serde_json::json;
+
+pub fn module() -> Module {
+    Module::new("Payments", |_ctx| {
+        RuntimeGroupSpec::new(GroupSpec::new("payments", "Manage payment methods"))
+            .with_command(add_command())
+    })
+}
+
+fn add_command() -> RuntimeCommandSpec {
+    RuntimeCommandSpec::new_with_context(
+        CommandSpec::new("add", "Add a payment method to your GoDaddy account (opens browser)")
+            .with_long(
+                "Opens your default browser to the GoDaddy payment methods management page.\n\
+                 Note: only credit card or Good-as-Gold can be used for domain purchases.",
+            )
+            .with_system("payments")
+            .with_tier(Tier::Mutate)
+            .no_auth(true),
+        |ctx| async move {
+            let url = account_url_for_env(&ctx.middleware.env);
+            open::that(&url)
+                .map_err(|e| CliCoreError::message(format!("failed to open browser: {e}")))?;
+            Ok(CommandResult::new(json!({
+                "info": "Browser opened to the payment methods management page. \
+                         Only credit card or Good-as-Gold can be used for domain purchases."
+            })))
+        },
+    )
+}
+
+fn account_url_for_env(env: &str) -> String {
+    let host = if env == "prod" {
+        "account.godaddy.com".to_owned()
+    } else {
+        format!("account.{env}-godaddy.com")
+    };
+    format!("https://{host}/payment-methods/add-payment?plid=1")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn prod_uses_bare_godaddy_domain() {
+        assert_eq!(
+            account_url_for_env("prod"),
+            "https://account.godaddy.com/payment-methods/add-payment?plid=1"
+        );
+    }
+
+    #[test]
+    fn ote_inserts_env_prefix() {
+        assert_eq!(
+            account_url_for_env("ote"),
+            "https://account.ote-godaddy.com/payment-methods/add-payment?plid=1"
+        );
+    }
+
+    #[test]
+    fn dev_inserts_env_prefix() {
+        assert_eq!(
+            account_url_for_env("dev"),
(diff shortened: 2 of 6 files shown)


rust/Cargo.lock
```diff
@@ -1105,6 +1105,7 @@ dependencies = [
  "domains-client",
  "fancy-regex",
  "httpmock",
+ "open",
```
sswaminathan-godaddy, line 1108: ps: this is evidently a cross platform "crate" that knows how to open browsers

rust/src/payments/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+                "Opens your default browser to the GoDaddy payment methods management page.\n\
+                 Note: only credit card or Good-as-Gold can be used for domain purchases.",
+            )
+            .with_system("payments")
+            .with_tier(Tier::Mutate)
+            .no_auth(true),
+        |ctx| async move {
+            let url = account_url_for_env(&ctx.middleware.env);
+            open::that(&url)
+                .map_err(|e| CliCoreError::message(format!("failed to open browser: {e}")))?;
```
jpage-godaddy: In case opening the browser is not working, perhaps we should print out the URL so the user can manually follow the URL. Or maybe that URL can be in the result message.

rust/Cargo.lock
jbrooks2-godaddy: cli-engine uses it too: https://github.com/godaddy/cli-engine/blob/main/Cargo.toml#L23. Used in the pkce auth flow: https://github.com/godaddy/cli-engine/blob/main/src/auth/pkce.rs#L506

rust/src/payments/mod.rs
jbrooks2-godaddy: Maybe follow the same shape as what oauth/pkce now does, and always emit? https://github.com/godaddy/cli-engine/blob/main/src/auth/pkce.rs#L652-L659

rust/src/payments/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+            Ok(CommandResult::new(json!({
+                "info": "Browser opened to the payment methods management page. \
+                         Only credit card or Good-as-Gold can be used for domain purchases."
+            })))
+        },
+    )
+}
+
+fn account_url_for_env(env: &str) -> String {
```
jbrooks2-godaddy: Should this use the same environment mechanics as the auth/domains base urls? https://github.com/godaddy/cli/blob/rust-port/rust/src/environments/mod.rs#L67-L83

rust/src/payments/mod.rs
sswaminathan-godaddy: I was hoping to not reveal the url in the result - do we expect failure in opening the browser?

I'll check https://github.com/godaddy/cli-engine/blob/main/src/auth/pkce.rs#L652-L659

rust/src/payments/mod.rs
sswaminathan-godaddy: will check

rust/src/payments/mod.rs
jbrooks2-godaddy: Yes, for anyone trying out the cli in a headless environment the browser-open would fail. We will eventually add the device-code flow for auth to make that work without the localhost callback; for this it's helpful to show the url so that they can browse there on a different device if needed.

sswaminathan-godaddy: I'm moderately concerned about exposing the URL that would allow brand new customers to add cards without going through a typical purchase, especially to llms. The information does get out if someone observes what it's opening on a browser. Headless environment feels like an edge case.

Let me know if you think we should still add it in the error response and I will update the pr

jbrooks2-godaddy: I don't think that this exposes much more. It's already - 
- In the source code here, on the public github
- In the browser that gets opened

I expect that we're going to make some reference to this on the docs pages too: "You need to have a credit card payment method set up to purchase domains via API, go here to set it up". Outputting to stderr doesn't feel like much of a lift from those three.

This is going to be out there, we'll need to spend the time protecting it regardless.

sswaminathan-godaddy: Yeah, agree. Will update

sswaminathan-godaddy: I'm chasing a risk call placement for this payment API. I guess we will acknowledge the gap for MVP

sswaminathan-godaddy: Done — on browser-open failure we now emit the URL to stderr (mirroring the pkce pattern: `drop(writeln!(stderr, ...))`) and return an error. On success the URL stays out of the output.

sswaminathan-godaddy: Followed the pkce pattern — on failure we lock stderr and `drop(writeln!)` the URL before returning the error. Emitting only on failure rather than always since opening a browser without prompting is already a surprise; the URL showing up only when it fails keeps the happy path clean.

rust/src/payments/mod.rs
sswaminathan-godaddy: Done — `account_url` is now a field on `ResolvedEnv`, following the `domains_api_url` pattern exactly. Derives to `account.godaddy.com` for prod and `account.{env}-godaddy.com` for others; overridable via `<ENV>_ACCOUNT_URL` env var or local `environments.toml`. The payments command now calls `environments::resolve()` instead of a hardcoded helper.

jbrooks2-godaddy (approved):

PR #65, opened 2026-06-16
feat: DNS record management + domain list, with output-field discovery

jpage-godaddy:
## What & why

Adds DNS record management and domain listing to the CLI, plus field discoverability across commands. All backed by the existing spec-generated `domains-client` crate, extended with the relevant operations of the same v1 Domains API.

### New commands
- **`gddy dns list|add|set|delete`** — manage a domain's DNS records.
  - `--type` is case-insensitive; `--data` is repeatable; `set`/`delete` are `Destructive` (preview with `--dry-run`).
  - `delete` rejects `NS`/`SOA` (registry/GoDaddy-managed — not deletable via the records API) with a clear message instead of an opaque conversion error.
- **`gddy domain list [--status …]`** — list the domains in your account.

### Command restructure
- **`api list` → `api domain list` + `api endpoint list --domain …`** *(refined during review).* The old `api list` emitted two different shapes (API-domain summaries vs endpoints-by-domain) under a single `default_fields`, which projected the endpoint shape down to empty `{}` objects (latent in JSON; surfaced in the human table once cli-engine 0.3.0 honors `default_fields`). Split into two commands, each with one consistent shape, `default_fields`, and output schema.

### `domains-client`
- Extended the trimmed OpenAPI spec to include the DNS-record operations (synthesizing the list-all / by-type GETs the published spec omits) and `GET /v1/domains`.
- Synthesized GETs preserve the optional `X-Shopper-Id` header so the list-all / by-type builders expose it like the documented `recordGet` *(refined during review)*.
- Made read-only response types tolerant of GoDaddy's sparse payloads (it omits spec-`required` fields like `contactRegistrant`/`renewDeadline` and returns `nameServers: null`), and stripped `date-time`/`uuid` formats so the lean reqwest/rustls dependency stack is unchanged.
- Derive `schemars::JsonSchema` on the generated types (pinned to cli-engine's schemars) for `--schema`.

### Field discovery
- Registered output schemas across `env`/`api`/`application`/`domain`/`dns` read **and** mutation commands — including the `dns add`/`set`/`delete` confirmation outputs *(added during review)* — via a small `output_schema!` macro, so `<cmd> --help` lists the available fields and `--schema` works. Genuinely passthrough/streaming commands (`api call`, `webhook events`, `actions describe`, `application deploy`) are intentionally left schema-less.

### Robustness
- `domain list` and `dns list` propagate JSON serialization errors as `CliCoreError` instead of silently masking them as `{}` *(refined during review)*.

### Engine
- Bump `cli-engine` to **0.3.0**: human output now honors `default_fields` (curated tables instead of every column; `--fields all` / `--output json` for everything), and `--schema` reports clearly when a command has no registered schema. The 0.3.0 breaking changes (human views no longer inferred; new `CommandSpec`/`MiddlewareRequest` fields) don't affect this crate.

## Verification
`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --workspace` all clean (gddy 164, domains-client 9, generate-api-catalog 8). Offline `httpmock` tests cover the record/list client ops (incl. sparse `nameServers: null` payloads); unit tests cover type validation, the NS/SOA delete guard, and list flag rules.

> Note: targets `rust-port` (this is the Rust-port lineage, not the legacy TS CLI on `main`).

### DIFF
--- a/rust/domains-client/scripts/regenerate-spec.sh
+++ b/rust/domains-client/scripts/regenerate-spec.sh
@@ -90,9 +90,13 @@ def synth_get(keep_param_names, op_id):
     op = copy.deepcopy(record_get)
     op["operationId"] = op_id
     op["produces"] = ["application/json"]
+    # Keep the named path params plus any header params (e.g. the optional
+    # X-Shopper-Id the record routes accept) so the synthesized ops mirror what
+    # the route supports; drop the query params (offset/limit) that only apply to
+    # the type+name GET.
     op["parameters"] = [
         prm for prm in op.get("parameters", [])
-        if prm.get("name") in keep_param_names
+        if prm.get("in") == "header" or prm.get("name") in keep_param_names
     ]
     return only_2xx(op)
 
--- a/rust/domains-client/scripts/regenerate-spec.sh
+++ b/rust/domains-client/scripts/regenerate-spec.sh
@@ -3,8 +3,10 @@
 #
 # Pipeline:
 #   1. Download the upstream GoDaddy Domains API spec (Swagger 2.0).
-#   2. Trim to the GET /v1/domains/available + GET /v1/domains/suggest operations
-#      and the transitive closure of definitions they reference.
+#   2. Trim to the GET /v1/domains (list owned domains), GET /v1/domains/available
+#      + GET /v1/domains/suggest operations, the /v1/domains/{domain}/records
+#      DNS-record operations, and the transitive closure of definitions they
+#      reference.
 #   3. Convert Swagger 2.0 -> OpenAPI 3.0 with `swagger2openapi` (Node, via npx).
 #
 # Run this ONLY when the upstream spec changes; the committed domains.oas3.json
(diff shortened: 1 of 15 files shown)


rust/src/domain/mod.rs
```diff
@@ -92,17 +104,71 @@ fn string_list(ctx: &CommandContext, key: &str) -> Vec<String> {
...
+                    .await
+                    .map_err(|e| CliCoreError::message(format!("listing domains failed: {e}")))?;
+
+                // Emit each summary as an object so `--fields`/default-field
+                // projection works (default shows domain/status/expires/renewAuto).
+                let domains: Vec<serde_json::Value> = resp
+                    .into_inner()
+                    .iter()
+                    .map(|d| serde_json::to_value(d).unwrap_or_else(|_| json!({})))
+                    .collect();
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in b1f5d37: `domain list` now propagates the `serde_json::to_value` error as a `CliCoreError` ("failed to serialize domain list") instead of emitting `{}`.

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,625 @@
...
+                                .into_inner()
+                        }
+                    };
+
+                    // Serialize each record to an object so `--fields` and the
+                    // default-field projection have `type`/`name`/`data`/`ttl`.
+                    let out: Vec<Value> = records
+                        .iter()
+                        .map(|r| serde_json::to_value(r).unwrap_or_else(|_| json!({})))
+                        .collect();
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in b1f5d37: `dns list` now propagates the serialization error as a `CliCoreError` instead of masking it as `{}`.

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,625 @@
...
+            .with_command(RuntimeCommandSpec::new_with_context(
+                with_record_write_args(
+                    CommandSpec::new(
+                        "add",
+                        "Add DNS records to a domain (appends; non-destructive)",
+                    )
+                    .with_system("domain")
+                    .with_tier(Tier::Mutate)
+                    .with_default_fields("domain,type,name,records")
+                    .with_scopes(&[DOMAINS_DNS_UPDATE_SCOPE]),
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in b1f5d37: registered an output schema for `dns add` (domain/type/name/records/action), so `--help`/`--schema` now list its fields.

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,625 @@
...
+            .with_command(RuntimeCommandSpec::new_with_context(
+                with_record_write_args(
+                    CommandSpec::new(
+                        "set",
+                        "Replace all records for a type+name (destructive: overwrites existing)",
+                    )
+                    .with_system("domain")
+                    .with_tier(Tier::Destructive)
+                    .with_default_fields("domain,type,name,records")
+                    .with_scopes(&[DOMAINS_DNS_UPDATE_SCOPE]),
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in b1f5d37: `dns set` now registers the same output schema (domain/type/name/records/action).

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,625 @@
...
+            // --- delete -----------------------------------------------------
+            .with_command(RuntimeCommandSpec::new_with_context(
+                CommandSpec::new(
+                    "delete",
+                    "Delete all records for a type+name (destructive: removes existing)",
+                )
+                .with_system("domain")
+                .with_tier(Tier::Destructive)
+                .with_default_fields("domain,type,name,deleted")
+                .with_scopes(&[DOMAINS_DNS_UPDATE_SCOPE])
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in b1f5d37: `dns delete` now registers an output schema (domain/type/name/deleted).

rust/src/api_explorer/mod.rs
```diff
@@ -215,6 +246,7 @@ fn list_command() -> RuntimeCommandSpec {
             .with_tier(Tier::Read)
             .no_auth(true)
             .with_default_fields("domain,title,endpoints,baseUrl")
+            .with_output_schema::<ApiDomain>()
```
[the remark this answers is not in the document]
jpage-godaddy: Done in a3dcd1c — went with the split (your pick). `api list` is now two commands, each with one consistent shape + default_fields + schema: `api domain list` ({domain,title,endpoints,baseUrl}) and `api endpoint list --domain X` ({operationId,method,path,summary}, --domain required). Verified both return full objects in json/human (no more empty {}).

rust/domains-client/scripts/regenerate-spec.sh
```diff
@@ -21,29 +23,132 @@ oas3="$openapi_dir/domains.oas3.json"
...
+record_get = d["paths"]["/v1/domains/{domain}/records/{type}/{name}"]["get"]
+
+def synth_get(keep_param_names, op_id):
+    op = copy.deepcopy(record_get)
+    op["operationId"] = op_id
+    op["produces"] = ["application/json"]
+    op["parameters"] = [
+        prm for prm in op.get("parameters", [])
+        if prm.get("name") in keep_param_names
+    ]
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in fa856dd: `synth_get` now keeps header params (so X-Shopper-Id is preserved) and only drops the query params (offset/limit) that don't apply. Regenerated domains.oas3.json — `record_get_all`/`record_get_by_type` now expose `.x_shopper_id()` like `record_get`.

rust/src/api_explorer/mod.rs
```diff
@@ -201,58 +241,30 @@ pub fn module() -> Module {
             "api",
             "Explore and call GoDaddy API endpoints",
         ))
-        .with_command(list_command())
+        .with_group(
```
jbrooks2-godaddy, line 244: Nit: the skill should be updated with this new behavior - https://github.com/godaddy/cli/blob/main/.agents/skills/godaddy-cli/SKILL.md#api-discovery-and-requests

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,649 @@
...
+                        .value_name("DOMAIN")
+                        .required(true)
+                        .help("Domain whose records to delete (e.g. example.com)"),
+                )
+                .with_arg(
+                    clap::Arg::new("type")
+                        .long("type")
+                        .value_name("TYPE")
+                        .required(true)
+                        .help("Record type (one of A, AAAA, CNAME, MX, NS, SOA, SRV, TXT)"),
```
jbrooks2-godaddy:
```suggestion
                        .help("Record type (one of A, AAAA, CNAME, MX, SRV, TXT)"),
```

NS/SOA can't be deleted https://github.com/godaddy/cli/pull/65/changes#diff-5b067632d8d17955283eed41c569c7171c5132b6c58993c6500e01766fb8c1f8R76-R83

rust/src/dns/mod.rs
```diff
@@ -0,0 +1,649 @@
...
+    ctx.args
+        .get(key)
+        .and_then(|v| v.as_str())
+        .map(str::to_owned)
+}
+
+/// Validate a user-supplied record type case-insensitively, returning the
+/// canonical upper-case wire string (for the builders' `type_` setters) and the
+/// parsed [`types::DnsRecordType`] (for record bodies).
+fn parse_record_type(raw: &str) -> Result<(String, types::DnsRecordType)> {
```
jbrooks2-godaddy: Can this validation be moved into Clap?

DNS invalid inputs pass dry-run / auth gates before validation. Because cli-engine dry-run short-circuits before handlers,` --dry-run dns set ... --type BOGUS` and `--dry-run dns delete ... --type NS` both return success. Same pattern means `dns list example.com --name www` enters auth instead of failing locally with `--name requires --type`.

Suggestions (from Codex) on what the validator could look like:
```rust
fn dns_record_type_parser() -> impl clap::builder::TypedValueParser<Value = String> {
    clap::builder::ValueParser::new(|raw: &str| {
        let upper = raw.to_ascii_uppercase();
        match upper.as_str() {
            "A" | "AAAA" | "CNAME" | "MX" | "NS" | "SOA" | "SRV" | "TXT" => Ok(upper),
            _ => Err(format!("invalid record type; expected one of {RECORD_TYPES}")),
        }
    })
}

fn deletable_dns_record_type_parser() -> impl clap::builder::TypedValueParser<Value = String> {
    clap::builder::ValueParser::new(|raw: &str| {
        let upper = raw.to_ascii_uppercase();
        match upper.as_str() {
            "A" | "AAAA" | "CNAME" | "MX" | "SRV" | "TXT" => Ok(upper),
            "NS" | "SOA" => Err(
                "NS and SOA records can't be deleted; they are managed by GoDaddy".to_string(),
            ),
            _ => Err("invalid record type; expected one of A, AAAA, CNAME, MX, SRV, TXT".to_string()),
        }
    })
}
```

And the arg parse:
```
clap::Arg::new("type")
    .long("type")
    .value_name("TYPE")
    .required(true)
    .value_parser(dns_record_type_parser())
    .help("Record type (one of A, AAAA, CNAME, MX, NS, SOA, SRV, TXT)")
```

rust/src/domain/mod.rs
```diff
@@ -28,7 +28,19 @@ const USER_AGENT: &str = concat!("godaddy-cli/", env!("CARGO_PKG_VERSION"));
 /// Declared on the commands via [`CommandSpec::with_scopes`] so cli-engine's
 /// OAuth scope step-up mints a token carrying it. Ignored on the sso-key path
 /// (sso-key auth is unscoped).
-const DOMAINS_READ_SCOPE: &str = "domains.domain:read";
+///
+/// Shared with the `dns` module: DNS record *reads* live under this same scope.
+pub(crate) const DOMAINS_READ_SCOPE: &str = "domains.domain:read";
```
jbrooks2-godaddy, line 33: I wonder how much cross-pollination of scopes we'll have; maybe these should be more central?

At some point once we have more of the functionality we should feel out the ux of the step-up. It's nice to get only the scopes you need but I think it would be annoying to do the oauth flow three times in a row because you do a few different things. Will probably want something at the start to give user a few pre-packaged scope bundles.

jbrooks2-godaddy (commented):
Minor suggestions, I do think it's worth fixing the validation issue so that dry-run works correctly

rust/src/domain/mod.rs
jpage-godaddy: Yeah, for sure. We do have a configurable set of base scopes, so if we decide to change that, I'm game. One potentially safe set of defaults could maybe be to give all `*:read` scopes.

rust/src/api_explorer/mod.rs
jpage-godaddy: Good call — updated `.agents/skills/godaddy-cli/SKILL.md` in 1cf1d04: `api list` / `api list --domain` examples are now `api domain list` / `api endpoint list --domain`, and the compatibility reserved-token note lists `domain`/`endpoint` instead of `list`.

rust/src/dns/mod.rs
jpage-godaddy: Applied in 1cf1d04 — the `dns delete` `--type` help now lists only A, AAAA, CNAME, MX, SRV, TXT, and the new value parser rejects NS/SOA at parse time with the GoDaddy-managed reason.

rust/src/dns/mod.rs
jpage-godaddy: Done in 1cf1d04. Moved `--type` validation into clap value-parsers (`parse_type_arg`, and a delete-only `parse_deletable_type_arg` that rejects NS/SOA), and made `--name` require `--type` via clap `.requires("type")`. So validation now runs at parse time, before the --dry-run/auth short-circuits: `--dry-run dns delete … --type NS` and `--dry-run dns set … --type BOGUS` both fail with a clear message, and `dns list example.com --name www` fails locally with a required-`--type` error instead of entering auth. Handlers now trust the validated value. Added a regression test (`invalid_dns_input_is_rejected_before_auth_or_dry_run`).

rust/src/domain/mod.rs
jpage-godaddy: Agreed — as the surface grows this is worth revisiting: centralizing the scope constants and offering a few pre-packaged scope bundles up front would avoid making users run the OAuth step-up several times in a row. I'd like to keep that out of this PR's scope (it's a UX design pass of its own), but happy to file a follow-up ticket to feel out the step-up UX. For now the per-command scopes are minimal and correct.

jbrooks2-godaddy: Agree it's not in scope here!

jbrooks2-godaddy (approved):

PR #69, opened 2026-06-18
fix: move esbuild from devDependencies to dependencies

tduong-godaddy:
## Problem

The published npm package (@godaddy/cli versions 0.5.1, 0.5.2, 0.5.3) is broken for users who don't have `esbuild` installed locally. When running any command (including `godaddy --version`), users encounter:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'esbuild' imported from .../cli.js
```

## Root Cause

1. **Runtime dependency**: `esbuild` is imported in `src/services/extension/bundler.ts`
2. **Loaded at startup**: This import is transitively loaded through:
   - `src/cli-entry.ts` imports `applicationCommand`
   - `src/cli/commands/application.ts` imports from `core/applications`
   - `src/core/applications.ts` imports from `services/extension/bundler`
3. **Marked as external**: `build.mjs` marks esbuild as external (not bundled)
4. **Wrong dependency type**: `esbuild` was in `devDependencies`, not `dependencies`

Since `devDependencies` are not installed when users install the package from npm, the CLI fails immediately on any command.

## Why Some Users Could Use It

The CLI worked for users who happened to have `esbuild` installed:
- In their project's `node_modules` (for their own build process)
- Installed globally

Node.js module resolution finds esbuild in parent directories, masking the packaging bug.

## Solution

Move `esbuild` from `devDependencies` to `dependencies` so it's always installed with the CLI package.

## Testing

Before fix (with esbuild not available):
```bash
npm install -g @godaddy/cli@0.5.3
godaddy --version
# Error: Cannot find package 'esbuild'
```

After fix:
```bash
npm install -g esbuild  # Install esbuild to test workaround
godaddy --version
# Works correctly
```

With this PR, esbuild will be installed automatically with the CLI.

## Impact

This fix will allow all users to use the CLI without requiring a separate esbuild installation, fixing the broken 0.5.x releases.


llessa-godaddy (commented):
Needs changeset. Otherwise PR is ✅

PR #77, opened 2026-07-02
Activate release during application deploy

dsingh5-godaddy:
Call activateRelease before setting application status to ACTIVE in godaddy application deploy. Add GraphQL mutation wrapper, MSW handler, and deploy tests asserting activation order.

Since we merged DEVX-363: Use latest ACTIVE release by activatedAt for app enablement, releases created via the CLI can no longer be enabled because the CLI doesn't activate releases yet.

I've raised a follow-up PR to:

1. Add bundles/uploads extensions 
2. Activate the release (added new)
3. Set the application as active

### DIFF
--- a/src/core/applications.ts
+++ b/src/core/applications.ts
@@ -14,6 +14,7 @@ import {
   archiveApplicationEffect as archiveAppServiceEffect,
   createApplicationEffect as createAppServiceEffect,
   createReleaseEffect as createReleaseServiceEffect,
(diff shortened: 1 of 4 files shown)


wcole1-godaddy (approved):
Approved. I reviewed the activation sequencing and ran validation in a clean detached worktree: typecheck, build, targeted deploy/CLI tests, application service integration test, diff check, and the Effect migration grep checks were clean.

PR #94, opened 2026-07-10
Hide cancelled/non-visible domains by default in domain list

jpage-godaddy:
## Summary
- `gddy domain list` now defaults to `statusGroups=VISIBLE`, hiding cancelled/confiscated/other non-visible domains instead of returning every status.
- Added `--show-hidden` to opt back into the old "show everything" behavior.
- `--status` is unchanged and still takes precedence over the default filter (e.g. `--status CANCELLED` still works).

## Why
Cancelled domains were cluttering `domain list` output. The v1 API's `statuses` enum has ~150 granular values (many internal registry workflow states), so hardcoding "all statuses minus CANCELLED" would be brittle. Instead this uses the API's existing `statusGroups=VISIBLE` grouping, which is maintained server-side.

## Test plan
- [x] `cargo test domain::list` — new `wants_visible_only` unit test + existing `parse_statuses` test pass
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo run -- domain list --help` — confirmed updated help text and new `--show-hidden` flag render correctly

### DIFF
--- a/rust/src/domain/list.rs
+++ b/rust/src/domain/list.rs
@@ -22,13 +22,23 @@ fn parse_statuses(raw: &[String]) -> Result<Vec<types::ListStatusesItem>> {
         .collect()
 }
 
+/// Whether the request should be scoped to the `VISIBLE` status group —
+/// GoDaddy's default view that hides cancelled/confiscated/other
+/// non-visible domains. Skipped when the caller passed an explicit
+/// `--status` filter or asked to see hidden domains via `--show-hidden`.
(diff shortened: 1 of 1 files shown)


jpage-godaddy:
Retargeting base branch to rust-port; reopening to trigger CI.

axburgess-godaddy (approved):

mmicco-godaddy (approved):

PR #97, opened 2026-07-13
fix(api-call): apply headers, error on non-2xx, GraphQL errors, output shape (DEVX-546)

smukherjee-godaddy:
## Problem

`gddy api call` had five defects that made it unreliable — and left the GraphQL commerce domains (`catalog-products`, `taxes`) impossible to call at all:

1. **Headers dropped** — `--header KEY:VALUE` was parsed but never attached to the request. Endpoints that require a header (the GraphQL commerce domains need `x-store-id`) could not be called.
2. **Non-2xx returned as success** — any non-2xx response was returned as a successful command result carrying the error body, so scripts/agents saw "success" on a 4xx/5xx.
3. **Non-JSON coerced to null** — a plain-text/HTML error body was silently turned into `null`, discarding the detail.
4. **GraphQL errors swallowed** — a GraphQL `200` response with a top-level `errors` array was reported as success.
5. **Output shape drift** — result was `{ status, body }` instead of the documented `{ endpoint, method, status, status_text, data }`.

## Fix

- **Headers:** apply every `--header` to the request (split on the first colon so values may contain colons; trimmed).
- **Non-2xx:** return an error (non-zero exit) with the status line + truncated response body, instead of a success envelope.
- **Non-JSON:** preserve the raw UTF-8 text; only truly empty/binary bodies become `null`.
- **GraphQL:** detect the GraphQL commerce surfaces by path (`graphql`/`subgraph`) and surface a non-empty `errors[]` as an error.
- **Output shape:** `{ endpoint, method, status, status_text, data }` (+ `headers` when `--include`).

Extracted `split_header`, `parse_response_body`, and `graphql_errors` as pure helpers with unit tests. The existing 403-with-required-scopes hint is preserved ahead of the generic non-2xx handler.

## Impact

- The GraphQL commerce domains (`catalog-products`, `taxes`) are now callable via `--header x-store-id: <uuid>`.
- Non-2xx and GraphQL errors fail loudly instead of masquerading as success.
- Response detail is never silently lost.

## Scope / dependency

This fixes correctness **at the `api call` layer**. The final envelope wrapping — success as `{ ok, command, result, next_actions }` on stdout and errors on stdout — is the output-contract work (DEVX-542 / DEVX-543); this PR composes with it and does not duplicate it.

Companion to DEVX-548 (catalog `$ref` resolution): that made the commerce schemas resolvable; this makes the endpoints callable.

## Testing

- `cargo test api_explorer` — 8 pass (3 new: `split_header_*`, `parse_response_body_*`, `graphql_errors_*`)
- `cargo clippy --bin gddy -- -D warnings` — clean
- `cargo fmt --check` — clean

Not covered here: a full `httpmock` + mock-credential integration test of the success path (`api call` requires a token) — that's the T4 test in the test-layer ticket.

### DIFF
--- a/rust/src/api_explorer/mod.rs
+++ b/rust/src/api_explorer/mod.rs
@@ -691,10 +691,7 @@ fn call_command() -> RuntimeCommandSpec {
                 .bearer_auth(&token)
                 .header("x-request-id", uuid::Uuid::new_v4().to_string());
 
-            // Apply user-supplied `--header KEY:VALUE` values (repeatable). These
-            // were parsed but never attached to the request, so endpoints that
-            // require a header — e.g. the GraphQL commerce domains need `x-store-id`
-            // — could not be called at all.
+            // Apply user-supplied `--header KEY:VALUE` values (repeatable).
             for h in string_list(&ctx.args, "header") {
                 let (key, val) = split_header(&h).ok_or_else(|| {
                     cli_engine::CliCoreError::message(format!(
--- a/rust/src/api_explorer/mod.rs
+++ b/rust/src/api_explorer/mod.rs
@@ -196,6 +196,34 @@ fn string_list(args: &serde_json::Map<String, Value>, key: &str) -> Vec<String> 
     }
 }
 
+/// Split a `--header` value of the form `KEY:VALUE` into trimmed parts.
+/// Splits on the first colon only, so values may themselves contain colons
+/// (e.g. a URL). Returns None when there is no colon.
+fn split_header(raw: &str) -> Option<(&str, &str)> {
+    raw.split_once(':').map(|(k, v)| (k.trim(), v.trim()))
+}
+
+/// Parse a response body as JSON when possible, otherwise preserve it as raw
+/// UTF-8 text. A non-JSON body (plain text / HTML error page) must not be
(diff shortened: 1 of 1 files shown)


rust/src/api_explorer/mod.rs
```diff
@@ -663,6 +691,19 @@ fn call_command() -> RuntimeCommandSpec {
                 .bearer_auth(&token)
                 .header("x-request-id", uuid::Uuid::new_v4().to_string());
 
+            // Apply user-supplied `--header KEY:VALUE` values (repeatable). These
+            // were parsed but never attached to the request, so endpoints that
+            // require a header — e.g. the GraphQL commerce domains need `x-store-id`
+            // — could not be called at all.
```
jpage-godaddy: No need to reference prior bugs; not useful for future readers.

```suggestion
            // Apply user-supplied `--header KEY:VALUE` values (repeatable). 
```

jpage-godaddy (approved):
Just one nit here and the rest looks good.

rust/src/api_explorer/mod.rs
smukherjee-godaddy: Updated in 68ec959. The comment now documents only the current header application behavior.

PR #104, opened 2026-07-14
Add Claude Code plugin marketplace and agent skill for the gddy CLI

dkoopman-godaddy:
## Summary

Adds a Claude Code plugin marketplace (`.claude-plugin/marketplace.json`) containing a skill for installing and using the `gddy` CLI.

## Test plan

- [x] `claude plugin validate .` and `claude plugin validate ./plugins/gddy` both pass with no warnings
- [x] Installed via `claude plugin marketplace add` + `claude plugin install gddy@godaddy` from the fork/branch — plugin and skill discoverable and installable
- [x] Installed via `npx skills add <tree-url> --skill gddy --agent claude-code` from the fork/branch
- [x] Asked the installed agent to check domain availability and add a DNS record — got the correct workflow reflecting SKILL.md guidance


.agents/skills/godaddy-cli/SKILL.md
```diff
@@ -370,7 +372,7 @@ godaddy application enable my-app --store-id <storeId>   # 8. Enable
 ```bash
 godaddy application info my-app                          # 1. Check current state
 godaddy application update my-app --description "New"    # 2. Update
-godaddy application validate my-app                      # 3. Validate
+godaddy application validate my-app                       # 3. Validate
```
dkoopman-godaddy:
```suggestion
godaddy application validate my-app                      # 3. Validate
```

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,84 @@
+# GoDaddy Domains API (raw HTTP fallback)
+
+Use this when the `gddy` CLI isn't installed or isn't a good fit for the task (CI pipelines, non-shell languages, etc.). When `gddy` is available, prefer it — it handles auth, quote-token bookkeeping, and idempotency for you.
+
+Base URL: `https://api.godaddy.com`
+
+Auth header: `Authorization: Bearer $GODADDY_PAT` (PAT generated from the Personal Access Token page under `https://developer.godaddy.com/docs/api-users/auth`). Legacy `sso-key key:secret` credentials still work for v1/v2 endpoints but not v3, and are deprecated — prefer a PAT for anything new.
+
+## Search and availability
```
jpage-godaddy: My worry about putting a bunch of specifics in here about the API is that we plan on covering all of GoDaddy's surface area eventually, and that could really cause the content here to bloat. I would much prefer if we could tell the agent to go to the portal docs instead; that way we don't have to update the skill whenever new feature are added or major changes are made.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+
+# gddy — GoDaddy domains & DNS CLI
+
+`gddy` is GoDaddy's beta CLI for domain search, registration, and DNS management. It's a separate, independently-installed tool from GoDaddy's older `godaddy` CLI (applications, deployments, webhooks) — different binary, different install method, different auth and config. Installing or using one doesn't affect the other. For application, deployment, or webhook tasks, use the `godaddy-cli` skill instead of this one.
+
+Both tools exist side by side because `gddy` is the newer of the two and under active development, with more surface area on its roadmap than what's documented here — it already has early, not-yet-stable support for things like applications and webhooks alongside its domain/DNS focus. Stick to the domain/DNS commands documented in this skill unless you've confirmed a newer capability is ready via `gddy --help`.
+
+It's under active development, and moves faster than its own docs and README — see "When the CLI and the docs disagree" below.
+
+## Fetching GoDaddy developer docs: curl the llms.mdx mirror, don't WebFetch the HTML
```
jpage-godaddy: Personally, I wouldn't reference `llms.mdx`. Just always use the help content that's in the CLI.

jpage-godaddy: Seems like this section isn't really CLI-related besides how to get the install instructions.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+irm https://github.com/godaddy/cli/releases/latest/download/install.ps1 | iex
+```
+
+Verify with `gddy --version`. This installs the `gddy` binary alongside any existing `godaddy` CLI — the two are separate tools.
+
+Authenticate with `gddy auth login` (opens a browser for OAuth). Check state with `gddy auth status`, sign out with `gddy auth logout`. For non-interactive use (CI, scripts), use a Personal Access Token instead — manage one with `gddy pat add/list/remove`, or set the `GDDY_PAT`/`GDDY_PAT_<ENV>` env var directly (checked before OAuth). Note PATs can't do everything: domain purchase specifically requires a customer-scoped OAuth login (see below).
+
+Every command accepts `--env <ote|prod>` to pick environment (default `prod`) and `--debug` for verbose output.
+
+## Domain purchase: it's a two-step quote-then-purchase flow, not one command
```
jpage-godaddy: I would put these details in the embedded CLI guide so the docs stay closer to the code. Either the CLI should be able to discover the guide itself or we can put in the skill that it should check `gddy guide domain-purchase`. Personally, I think we should keep the skill slim, just give hints on how it can find the docs.

.agents/skills/godaddy-cli/README.md
```diff
@@ -0,0 +1,30 @@
+# godaddy-cli skill
```
jpage-godaddy: Let's not bother with this CLI; the two will merge, and we don't currently have any external users of the CLI. In fact, maybe we shouldn't take up tokens mentioning it.

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,84 @@
+# GoDaddy Domains API (raw HTTP fallback)
+
+Use this when the `gddy` CLI isn't installed or isn't a good fit for the task (CI pipelines, non-shell languages, etc.). When `gddy` is available, prefer it — it handles auth, quote-token bookkeeping, and idempotency for you.
```
jpage-godaddy: Might this discourage it from following the install instructions for the CLI? Because installing the skill doesn't have the ability to install the CLI, we'll need some kind of prompt to tell us how to get started.

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,84 @@
...
+
+Base URL: `https://api.godaddy.com`
```
jpage-godaddy, line 5: Do we have a public URL for the specs? That would be the ideal since it should be able to figure out what to do just by reading/searching that.

.agents/skills/godaddy-cli/SKILL.md
dkoopman-godaddy: Fixed: applied the suggestion directly (also applied via the GitHub UI in the meantime — same change). See ba57596.

.agents/skills/gddy/reference/api.md
dkoopman-godaddy: Fixed: trimmed this file down to base URL/auth plus the two OpenAPI spec files (domains-v1.json, domains-v3.json) and two durable gotchas (idempotency-key retry safety, MCP read-only/no-auth scope) — dropped all the per-endpoint curl examples so this file shouldn't need touching as the API grows. See ba57596.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: agreed, removed the llms.mdx section from SKILL.md entirely. gddy --help is self-documenting for CLI usage, and the Setup section (moved up in the file) already has install instructions inline, so there's no docs-fetch needed there at all. See ba57596.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: trimmed the hardcoded flag examples (--period 1 --privacy, --quote-token <token> --agree --confirm) and now point to gddy domain quote --help / gddy domain purchase --help for the authoritative flag list. Kept the one paragraph explaining why it's two calls, not one, since --help won't state that reasoning on its own. Not referencing a gddy guide domain-purchase subcommand since I couldn't confirm one exists today. See ba57596.

.agents/skills/gddy/reference/api.md
dkoopman-godaddy: Fixed: added a pointer back to SKILL.md's Setup section for installing gddy itself, right at the top of this file. See ba57596.

.agents/skills/gddy/reference/api.md
dkoopman-godaddy: Yes — verified live: https://developer.godaddy.com/openapi/domains-v3.json (search/availability/registration-quotes/registrations/DNS records) and https://developer.godaddy.com/openapi/domains-v1.json (listing, bulk availability, contacts, transfers, legacy purchase). Between the two they cover everything this file previously hardcoded. Fixed: replaced the curated docs-page pointers with links to these spec files. See ba57596.

.agents/skills/godaddy-cli/README.md
dkoopman-godaddy: Discussed offline — keeping the godaddy-cli skill in this PR. godaddy's functionality is already ported to gddy behind a feature flag, but there are gaps (flagged by the original implementor) blocking that merge, so dropping godaddy-cli now would reduce what customers can actually do until that lands. On the token-bloat point: I looked for cross-reference fat to trim between the two skills' docs, but the existing paragraphs (gddy/SKILL.md's intro, godaddy-cli/SKILL.md's intro, and the "What it doesn't cover" note in godaddy-cli/README.md) are already minimal and load-bearing — they're what lets an agent tell these two similarly-named CLIs apart and avoid using the wrong one. No further trim without breaking that disambiguation, so no code change here. Will revisit once gddy reaches parity and the two merge.

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,19 @@
+# GoDaddy Domains API (raw HTTP fallback)
+
+Use this when the `gddy` CLI isn't installed or isn't a good fit for the task (CI pipelines, non-shell languages, etc.). When `gddy` is available, prefer it — it handles auth, quote-token bookkeeping, and idempotency for you. To install `gddy` itself, see the Setup section in SKILL.md.
```
jpage-godaddy: Still seems to be saying if `gddy` isn't installed we should only call the API instead of asking the user to install the API or if the agent can install it for them.

jpage-godaddy:
```suggestion
Use this when the `gddy` CLI cannot be installed or if it isn't a good fit for the task (CI pipelines, non-shell languages, etc.). When `gddy` is available, prefer it. To install `gddy` itself, see the Setup section of the `godaddy-cli` skill.
```

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,19 @@
...
+## Endpoint reference
+
+GoDaddy publishes machine-readable OpenAPI specs — fetch and search these directly instead of relying on hardcoded examples here, so this file doesn't need updating as the API grows:
+
+- Search, availability, registration (quote → register), DNS records: `https://developer.godaddy.com/openapi/domains-v3.json`
+- Domain listing, bulk availability, contacts, transfers, legacy purchase: `https://developer.godaddy.com/openapi/domains-v1.json`
+
+## Gotchas that won't change even as the API grows
+
+- **Registration is two calls, never one** — mirrors the CLI's `gddy domain quote` / `gddy domain purchase` split. Lock a quote first; it returns a single-use `quoteToken` valid for roughly 10 minutes. Execute the registration against that token with an `Idempotency-Key` header — required, not optional. Retrying with a *new* key can register (and charge for) the domain twice; always reuse the same key when retrying the same logical attempt.
```
jpage-godaddy:
```suggestion
- **Domain registration is two calls, never one** — mirrors the CLI's `gddy domain quote` / `gddy domain purchase` split. Lock a quote first; it returns a single-use `quoteToken` valid for roughly 10 minutes. Execute the registration against that token with an `Idempotency-Key` header — required, not optional. Retrying with a *new* key can register (and charge for) the domain twice; always reuse the same key when retrying the same logical attempt.
```

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
+---
+name: gddy
+description: Use GoDaddy's beta CLI (`gddy`) to search, register, and manage domains and DNS records. Load this skill whenever a task involves running `gddy` commands, parsing their JSON output, finding or buying a domain, or editing DNS records (A, CNAME, MX, TXT, etc.) for a domain hosted at GoDaddy. `gddy` is a separate tool from GoDaddy's older `godaddy` CLI (applications, deployments, webhooks) — do not use this skill for that tool, and don't trigger for other registrars or DNS providers (Cloudflare, Route 53, Namecheap, etc.) unless GoDaddy is explicitly involved.
+---
+
+# gddy — GoDaddy domains & DNS CLI
+
+`gddy` is GoDaddy's beta CLI for domain search, registration, and DNS management. It's a separate, independently-installed tool from GoDaddy's older `godaddy` CLI (applications, deployments, webhooks) — different binary, different install method, different auth and config. Installing or using one doesn't affect the other. For application, deployment, or webhook tasks, use the `godaddy-cli` skill instead of this one.
```
jpage-godaddy:
```suggestion
`gddy` is GoDaddy's beta CLI for domain search, registration, and DNS management. It's a separate from GoDaddy's older `godaddy` CLI. For application, deployment, or webhook tasks, use the `godaddy-cli` skill instead of this one.
```

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+Both tools exist side by side because `gddy` is the newer of the two and under active development, with more surface area on its roadmap than what's documented here — it already has early, not-yet-stable support for things like applications and webhooks alongside its domain/DNS focus. Stick to the domain/DNS commands documented in this skill unless you've confirmed a newer capability is ready via `gddy --help`.
+
+It's under active development, and moves faster than its own docs and README — see "When the CLI and the docs disagree" below.
```
jpage-godaddy: "moves faster than its own docs" is deceptive; it embeds its own docs. I assume this was in reference to the developer portal docs which are no-longer mentioned.

```suggestion
```

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+# macOS / Linux / Git Bash / MSYS2 / Cygwin
+curl -fsSL https://github.com/godaddy/cli/releases/latest/download/install.sh | bash
+
+# PowerShell
+irm https://github.com/godaddy/cli/releases/latest/download/install.ps1 | iex
+```
+
+Verify with `gddy --version`. This installs the `gddy` binary alongside any existing `godaddy` CLI — the two are separate tools.
```
jpage-godaddy: We already talked about the other binary at length, and the wording suggests it installs both.

```suggestion
Verify that installation works by running `gddy --version`.
```

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+```bash
+gddy domain suggest "coffee shop" --tlds com --tlds net --length-min 4 --length-max 15
+gddy domain available example.com --check-type fast   # or: full (live registry check, slower)
+gddy domain list --status ACTIVE
+gddy domain get example.com
+gddy domain agreements --tld com --privacy
+```
+
+Global `--limit <N>` caps result counts on list-like commands.
```
jpage-godaddy:
```suggestion
## Documentation

The `gddy` CLI is self-documenting. Use the following to get more information about its capabilities:

- The `--help` flag will give you detailed documentation on any command.
- The `gddy --search <keywords>` will help you find a command.
- The `gddy tree` command will give you a full listing of all commands.
- The `gddy guide` command lets you view detailed documentation that explains concepts and outlines complex workflows.

## Domain purchase

Domain purchases is a multi-step workflow. Run `gddy guide domain-purchase` for a detailed walkthrough. Always confirm with the user before finalizing a purchase — it's real money, and the action is not reversible.

File: .agents/skills/gddy/SKILL.md [hunk 14]
```diff
@@ -0,0 +1,86 @@
...
+
+## DNS management
```
jpage-godaddy, line 46: This is fine for now, but maybe it'd be better if we created a DNS guide which the agent can load on demand if DNS operations are relevant in the moment.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+```bash
+gddy dns list example.com --type A
+gddy dns add example.com --type A --name www --data 192.0.2.1 --ttl 3600
+gddy dns set example.com --type TXT --name @ --data "v=spf1 -all"
+gddy dns delete example.com --type A --name www
+```
+
+`add` appends a new record; `set` replaces every record matching that type+name; `delete` removes every record matching that type+name. Type-specific requirements: `MX`/`SRV` need `--priority`; `SRV` also needs `--port`, `--weight`, `--protocol`, `--service`; `CAA` requires `--tag` (`--flag` is optional).
+
+`set` and `delete` are destructive — support `--dry-run` to preview the change and `--reason <text>` for an audit trail. Run with `--dry-run` first and show the user what would change before applying it for real.
```
jpage-godaddy: The `--reason` text being required is buggy help text. It'll be fixed with [DEVEX-884](https://godaddy-corp.atlassian.net/browse/DEVEX-884)

```suggestion
`set` and `delete` are destructive. Run with `--dry-run` first and show the user what would change before applying it for real.
```

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Payments
+
+`gddy payments add` opens the browser to the account's payment-methods page — no card data is ever handled by the CLI itself. Purchases fail (403/422) without a valid payment method or Good-as-Gold balance on the account; check the error's `code` field, not just the HTTP status, to tell that apart from other failures.
```
jpage-godaddy: I don't think anyone outside of a select few _inside_ GoDaddy know what Good as Gold is.

plugins/godaddy-cli/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,8 @@
+{
```
jpage-godaddy: I'm still skeptical that we should publish this and then have to somehow retract it in a few months. This CLI is not used by customers now, and once we finalize the merging it'll never have been used by customers. It's only used by internal users dogfooding the systems to build our own in-house integrations.

.agents/skills/gddy/reference/api.md
dkoopman-godaddy: Fixed: reworded to "cannot be installed" (nudges toward attempting an install rather than assuming unavailability) and kept the "why prefer gddy" clause. One correction on the suggested text though — the Setup/install section for `gddy` lives in this skill's own SKILL.md, not `godaddy-cli`'s (that's the *other* CLI's skill). Kept the pointer as "see the Setup section in SKILL.md" to avoid sending readers to the wrong skill. See aa8c44d.

.agents/skills/gddy/reference/api.md
dkoopman-godaddy: Fixed: "Registration" → "Domain registration", and trimmed the explanation further — now points to `gddy guide domain-purchase` for the full walkthrough instead of restating it, keeping only the raw-HTTP-specific gotcha (Idempotency-Key requirement + reuse-on-retry). See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: trimmed to your suggested wording (with a grammar fix: "It's separate from" instead of "It's a separate from"). See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: deleted the "moves faster than its own docs" line. Kept the adjacent paragraph (line 10) since your comment didn't call that one out — but reworded it: dropped the "stick to documented commands unless..." framing and replaced it with a pointer to run `gddy --help` against the installed release for the authoritative, current command list. See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: trimmed to your suggested wording. See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: replaced with your suggested "Documentation" + "Domain purchase" sections, dropping the manually-maintained purchase/search walkthroughs in favor of pointing at the CLI's own --help/--search/tree/guide. Verified gddy --search, gddy tree, gddy guide, and gddy guide domain-purchase all exist and work as described against the installed 0.1.3 binary before accepting. Also fixed a dangling "(see below)" reference on line 26 that pointed at the now-removed OAuth detail — repointed it at `gddy guide domain-purchase`. See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Acknowledged — good idea, but there's no `gddy guide dns` today to point at, and authoring one is out of scope for this docs PR. Leaving the inline DNS section as-is for now; worth revisiting once/if that guide exists on the CLI side.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: dropped the `--reason <text>` audit-trail claim per your suggestion. Confirmed via `gddy dns delete --help` that its own text says "(required for destructive commands)" even though it isn't actually enforced — matches DEVEX-884. See aa8c44d.

.agents/skills/gddy/SKILL.md
dkoopman-godaddy: Fixed: replaced "Good-as-Gold balance" with "sufficient account balance" — drops the internal-only term without losing the actionable point (check the error's code field). See aa8c44d.

plugins/godaddy-cli/.claude-plugin/plugin.json
dkoopman-godaddy: @jpage-godaddy I assume you mean for the godaddy-cli maybe we should take this out and only publish the gddy skill?

We would still need a plugins/{name}/.claude-plugin/plugin.json file, we could name it gddy, "name": "gddy" and drop plugins/godaddy-cli/skills/godaddy-cli out of it.

We could take .agents/skills/godaddy-cli/SKILL.md back to unchanged, remove .agents/skills/godaddy-cli/README.md, remove mention of godaddy-cli in .agents/skills/gddy/ files, and change .claude-plugin/marketplace.json to name the plugin gddy (not godaddy-cli). The .agents/skills/godaddy-cli/README.md install instructions also need to change to claude plugin install gddy@godaddy

This is all very do-able. @jpage-godaddy give me a thumbs up if you want it and I'll do it.

dkoopman-godaddy: Done — dropped the godaddy-cli skill from the published plugin in dba8818. .agents/skills/godaddy-cli/SKILL.md is back to unchanged, its README is deleted, the plugin is renamed to gddy (plugins/gddy, marketplace entry, plugin.json), and the gddy skill/README no longer reference godaddy-cli.

jpage-godaddy (approved):

jpage-godaddy (approved):

.claude-plugin/marketplace.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "Skills for using GoDaddy's command-line tools with an AI coding agent.",
+  "plugins": [
+    {
+      "name": "gddy",
+      "source": "./plugins/gddy",
+      "description": "Skill for using GoDaddy's gddy CLI — domain search, registration, and DNS management.",
+      "author": {
+        "name": "GoDaddy"
+      }
+    }
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed the PR description

README.md
```diff
@@ -4,6 +4,12 @@ Agent-first CLI for interacting with GoDaddy Developer Platform.
 
 ## Installation
 
+### Using an AI coding agent?
+
+Point your agent at this README and ask it to install the `gddy` skill for domain search/registration/DNS. Once installed, the skill teaches the agent the CLI's JSON output contract and typical workflows, so it can drive the CLI correctly without guessing at flags. See [.agents/skills/gddy/README.md](./.agents/skills/gddy/README.md) for one-line install commands (Claude Code and 70+ other agents).
```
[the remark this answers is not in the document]
dkoopman-godaddy: No, no - we purposely decided to only promote and offer the gddy skill, skip this change and update the PR title and description to match what we landed on.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+Authenticate with `gddy auth login` (opens a browser for OAuth). Check state with `gddy auth status`, sign out with `gddy auth logout`. For non-interactive use (CI, scripts), use a Personal Access Token instead — manage one with `gddy pat add/list/remove`, or set the `GDDY_PAT`/`GDDY_PAT_<ENV>` env var directly (checked before OAuth). Note PATs can't do everything: domain purchase specifically requires a customer-scoped OAuth login (see `gddy guide domain-purchase`).
+
+Every command accepts `--env <ote|prod>` to pick environment (default `prod`) and `--debug` for verbose output.
+
+## Documentation
+
+The `gddy` CLI is self-documenting. Use the following to get more information about its capabilities:
+
+- The `--help` flag will give you detailed documentation on any command.
+- The `gddy --search <keywords>` will help you find a command.
```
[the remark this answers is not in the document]
dkoopman-godaddy: Pretty nit picky, but ok

.agents/skills/gddy/reference/api.md
```diff
@@ -0,0 +1,19 @@
+# GoDaddy Domains API (raw HTTP fallback)
+
+Use this when the `gddy` CLI cannot be installed or isn't a good fit for the task (CI pipelines, non-shell languages, etc.). When `gddy` is available, prefer it — it handles auth, quote-token bookkeeping, and idempotency for you. To install `gddy` itself, see the Setup section in SKILL.md.
+
+Base URL: `https://api.godaddy.com`
+
+Auth header: `Authorization: Bearer $GODADDY_PAT` (PAT generated from the Personal Access Token page under `https://developer.godaddy.com/docs/api-users/auth`). Legacy `sso-key key:secret` credentials still work for v1/v2 endpoints but not v3, and are deprecated — prefer a PAT for anything new.
```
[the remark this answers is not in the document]
dkoopman-godaddy: Good catch, please verify the gddy CLI really does check `GDDY_PAT/GDDY_PAT_<ENV>` then assuming it does, make this match.

README.md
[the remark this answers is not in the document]
dkoopman-godaddy: No, no - we purposely decided to only promote and offer the gddy skill, skip this change and update the PR title and description to match what we landed on.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,67 @@
+---
+name: gddy
+description: Use GoDaddy's beta CLI (`gddy`) to search, register, and manage domains and DNS records. Load this skill whenever a task involves running `gddy` commands, parsing their JSON output, finding or buying a domain, or editing DNS records (A, CNAME, MX, TXT, etc.) for a domain hosted at GoDaddy. `gddy` is a separate tool from GoDaddy's older `godaddy` CLI (applications, deployments, webhooks) — do not use this skill for that tool, and don't trigger for other registrars or DNS providers (Cloudflare, Route 53, Namecheap, etc.) unless GoDaddy is explicitly involved.
+---
```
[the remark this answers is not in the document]
dkoopman-godaddy: I agree, good catch

.agents/skills/gddy/reference/api.md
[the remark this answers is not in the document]
dkoopman-godaddy: Good catch, please verify the gddy CLI really does check GDDY_PAT/GDDY_PAT_<ENV> then assuming it does, make this match.

README.md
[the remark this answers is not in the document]
dkoopman-godaddy: Agreed — skipping this change. We intentionally scoped this PR to just the gddy skill; updated the PR title and description (Test plan) to reflect that instead of touching README.md.

.agents/skills/gddy/SKILL.md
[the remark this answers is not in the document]
dkoopman-godaddy: Fixed: removed the leading "The". See a8557c3.

.agents/skills/gddy/reference/api.md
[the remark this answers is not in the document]
dkoopman-godaddy: Fixed: verified against the installed gddy CLI (`gddy --help`, `gddy pat --help`) that it checks `GDDY_PAT`/`GDDY_PAT_<ENV>`, matching SKILL.md. Changed $GODADDY_PAT to $GDDY_PAT here. See a8557c3.

.agents/skills/gddy/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+The `gddy` CLI is self-documenting. Use the following to get more information about its capabilities:
+
+- The `--help` flag will give you detailed documentation on any command.
+- The `gddy --search <keywords>` will help you find a command.
+- The `gddy tree` command will give you a full listing of all commands.
+- The `gddy guide` command lets you view detailed documentation that explains concepts and outlines complex workflows.
+
+## Domain purchase
+
+Domain purchases is a multi-step workflow. Run `gddy guide domain-purchase` for a detailed walkthrough. Always confirm with the user before finalizing a purchase — it's real money, and the action is not reversible.
```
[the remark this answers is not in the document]
dkoopman-godaddy: Fixed via Jacob's commit eb385ab (same wording you suggested).

.agents/skills/gddy/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Any other AI coding agent
+
+This repo is also compatible with [skills](https://github.com/vercel-labs/skills), a package-manager-style installer for agent skills that isn't tied to Claude Code — it supports Cursor, Codex, Windsurf, opencode, and 70+ other agents in addition to Claude Code:
+
+```bash
+npx skills add godaddy/cli --skill gddy --agent claude-code
+```
+
+Swap `--agent claude-code` for whichever agent you use (`cursor`, `codex`, `windsurf`, `opencode`, ...). Run `npx skills add --help` for the full list of supported agents.
```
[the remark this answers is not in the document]
dkoopman-godaddy: Fixed: replaced with an <agent> placeholder (and updated the sentence below it that referenced --agent claude-code too). See a8557c3.

README.md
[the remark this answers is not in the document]
dkoopman-godaddy: Skipping — same call as before: we intentionally scoped this PR to just the gddy skill (godaddy-cli was dropped, see dba8818), so there's nothing to list it alongside. PR title/description already reflect the gddy-only scope.

.agents/skills/gddy/SKILL.md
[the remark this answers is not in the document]
dkoopman-godaddy: Fixed: added version/author/tags to match godaddy-cli/SKILL.md and cli-design/SKILL.md. See c9148e5.

.agents/skills/gddy/reference/api.md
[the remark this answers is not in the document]
dkoopman-godaddy: Already fixed by a8557c3, before this comment was posted — that commit changed $GODADDY_PAT to $GDDY_PAT here.

PR #124, opened 2026-07-17
fix(api-explorer): remove broken --query flag, add domains catalog source (DEVEX-898)

jpage-godaddy:
## Summary
- Removes `gddy api call`'s broken, hand-rolled `--query`/`-q` flag (DEVEX-898) — any lookup miss silently coerced the response to `null`. Points users at cli-engine's global `--expr`/`--filter` (JMESPath) flags instead, which are strictly more capable and already applied to every command's output.
- Wires `domains-client`'s vendored OpenAPI spec in as a new `domains` local catalog source for `api_explorer` (same mechanism already used for `hosting-nodejs`), so `gddy api domain/endpoint/describe/call` can browse and call the Domains registrar API for the first time — using the same spec that already drives the typed `domains-client` crate.
- Inverts `domains-client`'s dependency on `cli-engine` from pull to push: it now defines its own `TransportObserver` extension point, and the main crate pushes a `--debug transport` adapter into it, instead of `domains-client` calling into `cli-engine` directly. `domains-client` no longer depends on `cli-engine` at all.

## Test plan
- [x] `cargo build`, `cargo test --workspace`, `cargo clippy --workspace -- -D warnings` all clean
- [x] `gddy api call <endpoint> --query foo` now rejected by clap as an unknown flag
- [x] `gddy api call <endpoint> --expr ...` / `--output json` extraction verified against no-auth commands
- [x] `gddy api domain list` shows a new `domains` row; `gddy api endpoint list --domain domains` lists all 15 endpoints; `gddy api describe suggestDomains` shows full detail with correct scopes
- [x] `gddy domain suggest ... --debug transport` confirms the pushed `TransportObserver` adapter fires end-to-end with zero `cli-engine` dependency inside `domains-client` (`cargo tree -p domains-client` no longer lists `cli-engine`)

### DIFF
--- a/rust/api-catalog-sources.json
+++ b/rust/api-catalog-sources.json
@@ -86,14 +86,19 @@
     {
       "domain": "hosting-nodejs",
       "path": "schemas/openapi/hosting-nodejs-public-v1.yaml"
+    },
+    {
+      "domain": "domains",
+      "path": "domains-client/openapi/domains.oas3.json"
     }
   ],
   "legacyTypescript": {
     "status": "retired-on-rust-port",
     "sharedDomainCount": 20,
     "rustOnlyDomains": [
-      "hosting-nodejs"
+      "hosting-nodejs",
+      "domains"
     ],
-    "rationale": "The legacy TypeScript catalog contains the 20 remote domains above. hosting-nodejs is intentionally Rust-only because it is generated from the CLI's vendored public hosting specification."
+    "rationale": "The legacy TypeScript catalog contains the 20 remote domains above. hosting-nodejs and domains are intentionally Rust-only: each is generated from a vendored, manually-curated local spec (hosting-nodejs's public OpenAPI doc; domains' merged v1+v3 spec, which is also progenitor's source of truth for the typed domains-client crate)."
   }
 }
--- a/rust/domains-client/src/lib.rs
+++ b/rust/domains-client/src/lib.rs
@@ -42,8 +42,30 @@ pub enum BuildError {
     Http(#[from] reqwest::Error),
 }
 
-/// Bridges generated requests/responses into cli-engine's `--debug transport`
-/// logger.
+/// Observes every generated request/response, independent of any particular
+/// logging backend. This crate has no compile-time dependency on cli-engine
+/// (or any other framework) — the *caller* pushes an implementation in via
+/// [`set_transport_observer`] rather than this crate pulling one in.
+pub trait TransportObserver: Send + Sync {
+    fn on_request(&self, request: &reqwest::Request);
+    fn on_response(&self, status: reqwest::StatusCode, headers: &reqwest::header::HeaderMap);
+}
+
+static TRANSPORT_OBSERVER: std::sync::RwLock<Option<std::sync::Arc<dyn TransportObserver>>> =
+    std::sync::RwLock::new(None);
+
+/// Registers (or clears, with `None`) the process-wide transport observer.
+/// The main crate calls this with an adapter around its own logging
+/// framework — e.g. cli-engine's `--debug transport` bridge — before making
+/// any request through a [`Client`].
+pub fn set_transport_observer(observer: Option<std::sync::Arc<dyn TransportObserver>>) {
+    *TRANSPORT_OBSERVER
+        .write()
+        .expect("lock is never held across a panic") = observer;
+}
+
+/// Bridges generated requests/responses into the registered
+/// [`TransportObserver`], if any.
 ///
 /// progenitor generates every call as `client.pre(...)`, `client.exec(...)`,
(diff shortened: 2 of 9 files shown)


rust/domains-client/src/lib.rs
```diff
@@ -53,15 +75,21 @@ pub enum BuildError {
...
         _info: &progenitor_client::OperationInfo,
     ) -> Result<(), progenitor_client::Error<E>> {
-        cli_engine::transport::debug_log_reqwest_request(request);
+        if let Some(observer) = TRANSPORT_OBSERVER
+            .read()
+            .expect("lock is never held across a panic")
+            .as_ref()
+        {
+            observer.on_request(request);
+        }
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in a340482: cloned the `Arc` out and dropped the read guard before calling `observer.on_request(...)`, so the observer never runs while the lock is held.

rust/domains-client/src/lib.rs
```diff
@@ -70,12 +98,13 @@ impl progenitor_client::ClientHooks<()> for Client {
...
-                &[],
-            );
+        if let Ok(response) = result
+            && let Some(observer) = TRANSPORT_OBSERVER
+                .read()
+                .expect("lock is never held across a panic")
+                .as_ref()
+        {
+            observer.on_response(response.status(), response.headers());
         }
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in a340482: same fix — clone the `Arc` out and drop the guard before calling `observer.on_response(...)`.

rust/schemas/api/domains.json
```diff
@@ -0,0 +1,3000 @@
...
+      },
+      "title": "Simple Money",
+      "type": "object"
+    },
+    "uuid": {
+      "description": "A universally unique identifier (UUID) in [RFC-4122 format](https://tools.ietf.org/html/rfc4122).",
+      "type": "string"
+    }
+  },
+  "baseUrl": "https://api.ote-godaddy.com",
```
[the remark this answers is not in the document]
jpage-godaddy: Good catch on the confusion risk, but not fixing this one: `baseUrl` here comes straight from the vendored, upstream domains-client OpenAPI spec's `servers` block (`domains-client/openapi/domains.oas3.json`, refreshed via `domains-client/scripts/regenerate-spec.sh`) — hand-editing the generated catalog JSON would just get overwritten (and drift from the source of truth) on the next regen. It's cosmetic only: `api call` always resolves the real host from the active CLI environment via `api_url_for_env`, never from this stored field — same as the other 21 catalog domains. Worth fixing upstream in the domains spec/platform team's server list if it's misleading, but that's outside this PR's scope.

rust/src/domain/common.rs
```diff
@@ -14,6 +14,33 @@ fn map_env_err(e: environments::EnvError) -> CliCoreError {
...
+    }
+
+    fn on_response(&self, status: reqwest::StatusCode, headers: &reqwest::header::HeaderMap) {
+        cli_engine::transport::debug_log_reqwest_response(status, headers, &[]);
+    }
+}
+
+static TRANSPORT_OBSERVER_INIT: std::sync::Once = std::sync::Once::new();
+
+fn ensure_transport_observer_registered() {
```
qcai-godaddy, line 36: For any future generated client, I assume we’d reuse this same push/TransportObserver pattern?

rust/api-catalog-sources.json
```diff
@@ -86,14 +86,19 @@
     {
       "domain": "hosting-nodejs",
       "path": "schemas/openapi/hosting-nodejs-public-v1.yaml"
+    },
+    {
+      "domain": "domains",
+      "path": "domains-client/openapi/domains.oas3.json"
```
qcai-godaddy, line 92: After someone updates that OpenAPI, what’s the intended regen path for `gddy domain` and `gddy api`?

qcai-godaddy (approved):
Still ramping, but from what I followed this looks solid. Left a few questions just for my own understanding :)

rust/src/domain/common.rs
jpage-godaddy: Yep. At some point we might be able to extract something common in the engine, but for now we just have this one HTTP client.

rust/api-catalog-sources.json
jpage-godaddy: It's this https://github.com/godaddy/cli/blob/rust-port/rust/domains-client/scripts/regenerate-spec.sh script. I dislike how we now have three separate spec downloading/processing code paths. Created https://godaddy-corp.atlassian.net/browse/DEVEX-917 to track consolidating the logic.

PR #134, opened 2026-07-21
[DEVEX-708] fix(application validate): restore remote application state checks

qcai-godaddy:
## Summary
- Restore `application validate <name>` to fetch and check remote app state (parity with the TypeScript CLI), instead of only deserializing local `godaddy.toml`
- Report `{ valid, errors, warnings }`: missing URL → error; missing proxy URL / `INACTIVE` → warnings
- Update help text and next-actions; add unit tests for clap args and validation rules

## Test plan
- [x] `cargo build`
- [x] `cargo test -- application::commands::tests::validate`
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/application/commands/mod.rs
+++ b/rust/src/application/commands/mod.rs
@@ -505,7 +505,7 @@ fn init_command() -> RuntimeCommandSpec {
                     "Create the first release",
                 )
                 .with_param("application-id", required_value(app_id))
-                .with_param("version", required_value("0.0.1")),
+                .with_param("version", NextActionParam::required()),
             ]))
         },
     )
--- a/rust/src/application/commands/mod.rs
+++ b/rust/src/application/commands/mod.rs
@@ -5,7 +5,7 @@ use cli_engine::{
 use serde_json::{Value, json};
 
 use crate::application::client::{ApplicationClient, UploadOptions, api_url_for_env};
-use crate::next_action::next_action;
+use crate::next_action::{next_action, required_value};
 use crate::output_schema::output_schema;
 // App-registry mutations declare their scopes so `apps.app-registry:write` is
 // requested on demand (OAuth step-up), not granted at every login — it's a
@@ -157,7 +157,7 @@ fn deploy_result_event(
             "application info --name <name>",
             "Inspect the deployed application",
         )
-        .with_param("name", NextActionParam::value(name))],
+        .with_param("name", required_value(name))],
     })
 }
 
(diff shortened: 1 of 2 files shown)


rust/src/application/commands/mod.rs
```diff
@@ -504,40 +514,89 @@ fn init_command() -> RuntimeCommandSpec {
...
+                )
+                .with_param(
+                    "name",
+                    NextActionParam {
+                        value: Some(name),
+                        required: true,
+                        ..Default::default()
+                    },
+                ),
+            ]))
```
jpearlman-godaddy: Changes look good overall. Still getting the hang of Rust so would you mind explaining what this section of the change is doing? What is it doing differently from what was there before? Is it just using the new output_schema?

qcai-godaddy: Not quite — output_schema was already there and just declares the JSON output shape. The real change is what validate does:

Before: only read/parsed the local `godaddy.toml` file (no network, no auth) and returned { valid: true, path }.
After: `validate <name>` fetches the app from the remote platform (authenticated) and validates its actual state, matching the [TypeScript](https://github.com/godaddy/cli/blob/main/src/core/applications.ts#L708-L755) CLI.

`client.get_application(&name)` makes an authenticated GraphQL request to the `app-registry-subgraph`, returning `id`, `status`, `url`, `proxyUrl`, etc. That's why we removed `.no_auth(true)` and added the required name arg.

jpearlman-godaddy (approved):

rust/src/application/commands/mod.rs
```diff
@@ -492,8 +494,16 @@ fn init_command() -> RuntimeCommandSpec {
             });
             Ok(CommandResult::new(result).with_next_actions(vec![
                 next_action(
-                    "application validate",
-                    "Validate the generated godaddy.toml",
+                    format!("application validate {name}"),
+                    "Validate the remote application state",
+                )
+                .with_param(
```
jpage-godaddy: These `.with_param` functions fill in any angle-bracketed text with a value, so while this works, it's inconsistent with other uses. 

I know it's not in scope of your changes, but we probably ought to add a `.with_param(...)` for the other `next_action` so that the `<id>` can be filled in with the actual `id` used for the application that was validated.

rust/src/application/commands/mod.rs
```diff
@@ -504,40 +514,89 @@ fn init_command() -> RuntimeCommandSpec {
...
+
+            let app_id = app["id"].as_str().unwrap_or("").to_owned();
+            let (valid, errors, warnings) = validate_remote_application(app);
+            Ok(CommandResult::new(json!({
+                "valid": valid,
+                "errors": errors,
+                "warnings": warnings,
+            }))
+            .with_next_actions(vec![
+                next_action(
```
jpage-godaddy: Same feedback for these other `next_action` cases.

rust/src/application/commands/mod.rs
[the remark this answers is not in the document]
jpage-godaddy: Ah yes, Copilot noticed the same thing as me.

rust/src/application/commands/mod.rs
```diff
@@ -492,52 +492,102 @@ fn init_command() -> RuntimeCommandSpec {
...
+    if url.is_empty() {
+        errors.push("Application URL is required".to_owned());
+    }
+    let proxy_url = app["proxyUrl"].as_str().unwrap_or("");
+    if proxy_url.is_empty() {
+        warnings.push("Proxy URL is not set".to_owned());
+    }
+    if app["status"].as_str() == Some("INACTIVE") {
+        warnings.push("Application is currently inactive".to_owned());
+    }
```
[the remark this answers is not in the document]
qcai-godaddy: since those URLs come from the remote app (already enforced at init), validate doesn't need to re-check routability.

rust/src/application/commands/mod.rs
```diff
@@ -492,52 +496,100 @@ fn init_command() -> RuntimeCommandSpec {
...
+                )
+                .with_param("name", required_value(name)),
                 next_action(
-                    "application release --application-id <id> --version 0.0.1",
+                    "application release --application-id <application-id> --version <version>",
                     "Create the first release",
-                ),
+                )
+                .with_param("application-id", required_value(app_id))
+                .with_param("version", required_value("0.0.1")),
```
jpage-godaddy: Isn't `0.0.1` just an example? Let's leave `version` unspecified.

jpage-godaddy (approved):

PR #158, opened 2026-07-28
chore(cicd): code scanning / CI port is thin

mguerrero3-godaddy:
- Added a permissions block (contents: read, checks: write, issues: write) required by the audit action.
- Added an "Audit dependencies for known vulnerabilities" step using rustsec/audit-check@v2.0.0 (pinned by SHA), scoped to working-directory: rust.
- That step ignores 2 advisories via ignore: RUSTSEC-2026-0194,RUSTSEC-2026-0195 -> both against quick-xml, pulled in transitively through phonenumber's capped version range, with a comment explaining no fix is available until phonenumber bumps it.

Cargo audit now reports 0 vulnerabilities and only the 2 known-blocked quick-xml advisories ignored (dependency of dependency, and may be involved in internal API wiring with risk If we override it here)

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -11,6 +11,10 @@ jobs:
   cicd:
     runs-on: ubuntu-latest
     concurrency: ${{ github.workflow }}-${{ github.ref }}
+    permissions:
+      contents: read
+      checks: write
+      issues: write
     defaults:
       run:
         working-directory: rust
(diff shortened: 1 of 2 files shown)


jpearlman-godaddy:
Should this be merging into the rust-port branch rather than main?

mguerrero3-godaddy:
> Should this be merging into the rust-port branch rather than main?

@jpearlman-godaddy I don't see the rust-port branch anymore. I do see one rust-port branch saying "release-please" so I'm not certain that's It. Plus this main branch now has the rust code, so I think the port is already in place?

jpage-godaddy (approved):

jpage-godaddy:
(BTW, added you as a maintainer so you can work off branches now so we don't have to approve the running of workflows)

PR #164, opened 2026-07-30
fix(ci): authenticate catalog-drift's spec fetch via GitHub App token

jpage-godaddy:
## Summary
- `catalog-drift.yml` reads the gdcorp-platform spec repos to check for API catalog drift, but the token it depends on (`CATALOG_SPEC_TOKEN`) was never provisioned.
- Replaces that expectation with a GitHub App installed on `gdcorp-platform` (read-only Contents access). The workflow now exchanges the app's Client ID + private key (`SPEC_SYNC_CLIENT_ID` / `SPEC_SYNC_PRIVATE_KEY` secrets) for a short-lived installation token via `actions/create-github-app-token`, scoped explicitly to the `gdcorp-platform` owner.

## Test plan
- [x] Confirm `SPEC_SYNC_CLIENT_ID` / `SPEC_SYNC_PRIVATE_KEY` secrets are set on this repo and the app is installed on `gdcorp-platform` with Contents:Read
- [ ] Run the workflow via `workflow_dispatch` and confirm the token-generation step succeeds and the catalog regenerates without drift errors

### DIFF
--- a/.github/workflows/catalog-drift.yml
+++ b/.github/workflows/catalog-drift.yml
@@ -7,9 +7,11 @@ name: API catalog drift
 # list) lives in `cargo test`
 # (committed_catalog_matches_expected_domains_and_manifest) and runs on every PR.
 #
-# Requires a token with read access to the gdcorp-platform spec repos, provisioned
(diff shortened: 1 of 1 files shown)


runruh-godaddy (commented):
lgtm

PR #169, opened 2026-07-31
feat(api): restore agent-safe describe/list/search output [DEVEX-702]

jpage-godaddy:
## Summary

Restores agent-safe output for `api describe`/`domain list`/`endpoint list`/`search` (parity gap #9 from the original Rust-port review, DEVEX-702:

- **`api describe`**: schema summarization (`schema_type_label`/`summarize_schema`) instead of dumping raw, unresolved JSON schema; GraphQL operation summaries (wires up catalog data that was already embedded but silently dropped); a `--method`/`-m` filter; and fuzzy multi-match resolution (0 matches → error, 1 → transparent resolution, >1 → a `{message, matches}` response with per-match `next_actions`).
- **`api endpoint list` / `search`**: rows now include `scopes` and `graphqlOperations`; fixed a blind spot where GraphQL operation names weren't searchable at all. (`api domain list` is unchanged — it lists domains, not endpoints, so scopes/GraphQL counts don't apply there.)

## Explicitly deferred

No truncation or overflow-file mechanism is included — schemas and GraphQL operation lists are always returned in full, however large. We prototyped a `schema_detail`/`operationsDetail` overflow-file escape hatch during development and reverted it; it's messy (arbitrary temp files, no navigational structure) and not the right long-term shape. Redesigning this as command "hyperlinks" + pagination is tracked separately in DEVEX-967, linked from DEVEX-702.

The same applies to gaps 1, 5, 6, & 7 from the original review (bare-array output on `domain list`/`endpoint list`/`search` instead of a `{total, shown, truncated, full_output}` wrapper) — cli-engine's `--output human` table rendering and `--limit`/`--offset`/`--filter` pipeline only work on bare top-level arrays, so wrapping would silently break those for these three commands. Deferred to the same follow-up.

## Test plan

- [x] `cargo test` (445/445 passing)
- [x] `cargo clippy -- -D warnings` (clean)
- [x] `cargo fmt -- --check` (clean)
- [x] Manual verification: exact/fuzzy/multi-match describe resolution, `--method` filtering (including the documented quirk where it's ignored during fuzzy fallback), GraphQL summary on a 149-operation endpoint, `scopes`/`graphqlOperations` on list/search rows, GraphQL operation names now searchable, `fullPath` stays hostless across environments (prod and OTE)

### DIFF
--- a/rust/src/api_explorer/mod.rs
+++ b/rust/src/api_explorer/mod.rs
@@ -25,7 +25,7 @@ output_schema!(ApiEndpoint {
     "path": "string";
     "summary": "string", optional;
     "scopes": "[]string";
-    "graphql_operations": "number", optional;
+    "graphqlOperations": "number", optional;
 });
 
 // `api endpoint list --domain X` lists endpoints within one domain, so each row
@@ -36,7 +36,7 @@ output_schema!(ApiDomainEndpoint {
     "path": "string";
     "summary": "string", optional;
     "scopes": "[]string";
-    "graphql_operations": "number", optional;
+    "graphqlOperations": "number", optional;
 });
 
 output_schema!(ApiOperation {
@@ -347,30 +347,26 @@ fn search_endpoints<'a>(catalog: &'a [Domain], query: &str) -> Vec<(&'a Domain, 
         .flat_map(|domain| {
             let q = q.clone();
             domain.endpoints.iter().filter_map(move |ep| {
-                let graphql_searchable = ep
-                    .graphql
-                    .as_ref()
-                    .map(|g| {
-                        g.operations
-                            .iter()
-                            .map(|op| format!("{} {}", op.kind, op.name))
-                            .collect::<Vec<_>>()
-                            .join(" ")
-                    })
-                    .unwrap_or_default();
                 let haystack = format!(
-                    "{} {} {} {} {}",
+                    "{} {} {} {}",
                     ep.operation_id.to_lowercase(),
                     ep.path.to_lowercase(),
                     ep.summary.to_lowercase(),
                     ep.description.to_lowercase(),
-                    graphql_searchable.to_lowercase(),
                 );
-                if haystack.contains(&q) {
-                    Some((domain, ep))
-                } else {
-                    None
-                }
+                // Only scan GraphQL operations (up to 149 on some domains)
+                // when the core fields didn't already match, and stop at the
+                // first hit instead of concatenating every operation into one
+                // string up front.
+                let matches = haystack.contains(&q)
+                    || ep.graphql.as_ref().is_some_and(|g| {
+                        g.operations.iter().any(|op| {
+                            format!("{} {}", op.kind, op.name)
+                                .to_lowercase()
+                                .contains(&q)
+                        })
+                    });
+                if matches { Some((domain, ep)) } else { None }
             })
         })
         .collect()
@@ -742,7 +738,7 @@ fn endpoint_list_command() -> RuntimeCommandSpec {
                         "path": ep.path,
                         "summary": ep.summary,
                         "scopes": ep.scopes,
-                        "graphql_operations": ep.graphql.as_ref().map(|g| g.operation_count),
+                        "graphqlOperations": ep.graphql.as_ref().map(|g| g.operation_count),
                     })
                 })
                 .collect();
@@ -882,11 +878,18 @@ fn describe_command() -> RuntimeCommandSpec {
                     .collect()
             });
 
+            // Strip `base_url`'s scheme+host generically (not a hard-coded
+            // prod hostname) so `fullPath` stays a hostless path prefix +
+            // endpoint path consistently across every environment, not just
+            // prod — `resolve_catalog_base_url` rewrites the host per env
+            // (e.g. `api.ote-godaddy.com`), which a literal prod-host strip
+            // would silently leave un-stripped.
             let full_path = {
-                let raw = format!("{base_url}{}", ep.path);
-                raw.strip_prefix("https://api.godaddy.com")
-                    .map(str::to_owned)
-                    .unwrap_or(raw)
+                let without_scheme = base_url
+                    .split_once("://")
+                    .map_or(base_url.as_str(), |(_, rest)| rest);
+                let path_prefix = without_scheme.find('/').map_or("", |i| &without_scheme[i..]);
+                format!("{path_prefix}{}", ep.path)
             };
 
             Ok(CommandResult::new(json!({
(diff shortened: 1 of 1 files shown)


rust/src/api_explorer/mod.rs
```diff
@@ -282,12 +347,24 @@ fn search_endpoints<'a>(catalog: &'a [Domain], query: &str) -> Vec<(&'a Domain,
         .flat_map(|domain| {
             let q = q.clone();
             domain.endpoints.iter().filter_map(move |ep| {
+                let graphql_searchable = ep
+                    .graphql
+                    .as_ref()
+                    .map(|g| {
+                        g.operations
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in 8dc4cc4: search now checks the core fields first and only scans GraphQL operations when they didn't already match, short-circuiting on the first hit via `.any()` instead of concatenating every operation into one string up front.

rust/src/api_explorer/mod.rs
```diff
@@ -466,34 +778,131 @@ fn describe_command() -> RuntimeCommandSpec {
...
+                    })
+                    .collect()
+            });
+
+            let full_path = {
+                let raw = format!("{base_url}{}", ep.path);
+                raw.strip_prefix("https://api.godaddy.com")
+                    .map(str::to_owned)
+                    .unwrap_or(raw)
+            };
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in 8dc4cc4: strips base_url's scheme+host generically (split on `://` then find the first `/`) instead of a hard-coded prod hostname, so fullPath stays a hostless path in every environment. Added a regression test (describe_full_path_is_hostless_in_a_non_prod_env) covering OTE specifically, which was silently returning a full URL before this fix.

rust/src/api_explorer/mod.rs
```diff
@@ -24,6 +24,8 @@ output_schema!(ApiEndpoint {
     "method": "string";
     "path": "string";
     "summary": "string", optional;
+    "scopes": "[]string";
+    "graphql_operations": "number", optional;
 });
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in 8dc4cc4: renamed to graphqlOperations everywhere (output_schema declarations + JSON builders in endpoint_list_command/search_command).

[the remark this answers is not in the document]
jpage-godaddy: Correct catch — updated the PR description: the enrichment landed on endpoint list/search rows, not domain list (which only has domain-level fields).

rust/src/api_explorer/mod.rs
```diff
@@ -208,11 +250,34 @@ fn find_endpoint<'a>(catalog: &'a [Domain], query: &str) -> Option<(&'a Domain,
...
+    catalog: &'a [Domain],
+    query: &str,
+    method: Option<&str>,
+) -> Option<(&'a Domain, &'a Endpoint)> {
+    let q = query.to_lowercase();
+    catalog.iter().find_map(|domain| {
+        domain
+            .endpoints
+            .iter()
+            .find(|ep| {
```
qcai-godaddy: If method isn't provided and multiple endpoints share an exact path match, this finds first hit — should we return the multi-match payload instead?

rust/src/api_explorer/mod.rs
```diff
@@ -289,16 +354,257 @@ fn search_endpoints<'a>(catalog: &'a [Domain], query: &str) -> Vec<(&'a Domain,
...
+        Value::Null => String::new(),
+        Value::Bool(b) => b.to_string(),
+        Value::Number(n) => n.to_string(),
+        other => other.to_string(),
+    }
+}
+
+fn summarize_schema(schema: Option<&Value>) -> Option<Vec<SchemaSummaryProperty>> {
+    let obj = schema?.as_object()?;
+    let Some(properties) = obj.get("properties").and_then(Value::as_object) else {
```
qcai-godaddy, line 560: Checking the schema repo, most catalog request bodies are `$ref` only, but we only summarize `properties`/`type`/`enum`. Should we resolve `$ref` against `$defs` too?

rust/src/api_explorer/mod.rs
```diff
@@ -466,34 +774,138 @@ fn describe_command() -> RuntimeCommandSpec {
...
+                                })
+                                .collect();
+                            let next_actions = hits
+                                .iter()
+                                .map(|(_, ep)| {
+                                    next_action(
+                                        "api describe <endpoint>",
+                                        format!("{} {} — {}", ep.method, ep.path, ep.summary),
+                                    )
+                                    .with_param("endpoint", required_value(ep.path.clone()))
```
qcai-godaddy:
```suggestion
                                    .with_param("endpoint", required_value(ep.path.clone()))
                                    .with_param("method", required_value(ep.method.clone()))
```
Add method so when the same path appears more than once in matches next_actions suggest the method too?

rust/src/api_explorer/mod.rs
jpage-godaddy: _Great_ find. Will do.

rust/src/api_explorer/mod.rs
jpage-godaddy: Good catch — fixed in c2a221d. find_endpoint_exact now returns every exact match instead of just the first; if there's more than one (common — e.g. GET/POST sharing a collection path) it goes through the same multi_match_result() path as an ambiguous fuzzy match, rather than silently picking whichever endpoint it saw first. Also extended it to match a concrete path against a templated catalog path (path_matches_template), so this is now consistent with what api call already does via find_endpoint.

rust/src/api_explorer/mod.rs
jpage-godaddy: Implemented in c2a221d: added a Domain.defs field capturing the embedded $defs, and schema_type_label/summarize_schema now resolve $ref against it (bounded depth to guard against a def-to-def cycle), both at the top level and per-property, preferring a local override over the resolved def per JSON Schema's sibling-keyword rules. Confirmed the scope first — 63 of 82 catalog request bodies (77%) are a bare $ref with no inline properties, so this was returning null for the large majority of write endpoints. createBusiness (whose body is {"$ref": "#/$defs/Business"}) now resolves to the full 27-property list.

rust/src/api_explorer/mod.rs
jpage-godaddy: Applied (via a shared multi_match_result() helper now used by both the fuzzy and exact-collision paths, in c2a221d) — each next_action carries both endpoint and method params, and the command template is "api describe <endpoint> --method <method>".

rust/src/api_explorer/mod.rs
qcai-godaddy: nice!

qcai-godaddy (approved):

PR #174, opened 2026-08-01
fix(cli): drop experimental platform suggestion, fix stale --search doc

jpage-godaddy:
## Summary
- Remove `platform app list` from the root `next_actions` list — `platform` is gated behind `Stage::Experimental` and hidden at the default `Ga` min-stage, but this suggestion surfaced unconditionally for everyone.
- Fix the `gddy` skill docs (`SKILL.md`, `README.md`), which still referenced the removed `--search` global flag instead of the `search <query...>` subcommand introduced in cli-engine 0.5.0.

## Test plan
- [x] `cargo check`
- [x] `cargo test`
- [x] `cargo fmt --check`


jpage-godaddy:
Merging to avoid live bug.

PR #178, opened 2026-08-03
fix(install): fix PowerShell 5.1 compatibility bugs in install.ps1

jpage-godaddy:
## Summary
- Windows PowerShell 5.1 users hit `install.ps1` failures at two points:
  - Architecture detection used `[System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture`, which requires .NET Framework 4.7.1+; on older runtimes the property genuinely doesn't exist, so it threw `PropertyNotFoundException` before anything downloaded.
  - The PATH-membership check piped `Where-Object` straight into `.Count`; with 0 or 1 matches that pipeline collapses to `$null`/a bare string in PS 5.1 (no auto `Count`, unlike pwsh 7+), tripping `Set-StrictMode`.
- Replaced arch detection with the `PROCESSOR_ARCHITECTURE`/`PROCESSOR_ARCHITEW6432` env vars (no .NET type dependency) and wrapped the `Where-Object` pipeline in `@(...)` so `.Count` is always valid.

## Test plan
- [x] Reproduced both original errors on Windows PowerShell 5.1 via `irm .../install.ps1 | iex`
- [x] Verified the fixed script completes end-to-end: detects platform, downloads + verifies checksum, installs `gddy.exe`, and updates the user `PATH`
- [ ] CI / reviewer to confirm no other consumers of the old arch-detection logic

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -52,9 +52,15 @@ try { [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocol]::Tls1
 
 # ── 1. Detect platform ───────────────────────────────────────────────────────
 
-$arch = [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLower()
-switch ($arch) {
-    'x64'   { $platform = 'x86_64-pc-windows-msvc' }
+# Use the PROCESSOR_ARCHITECTURE env vars (not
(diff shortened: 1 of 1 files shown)


jpage-godaddy:
Going to go ahead and merge; I don't think many people have a Windows machine to confirm, so you'll all just have to trust me :D

PR #188, opened 2026-08-06
docs: Interactive domain purchase design

rts1-godaddy:
Proposal to support an Interactive domain purchase for cli users. This is intended for humans use only.


docs/design/INTERACTIVE_DOMAIN_WIZARD.md
```diff
@@ -0,0 +1,483 @@
...
+| Professional Email | Email API | `POST /v1/email/domains/{domain}` |
+| DNS Records | Domains v1 | `PATCH /v1/domains/{domain}/records` |
+| Website Builder | (future) | Placeholder for hosting product integration |
+
+## Command Design
+
+### Command Signature
+
+```
+gddy domain register [DOMAIN] [--interactive | --non-interactive]
```
jpage-godaddy, line 119: I wonder if we should have multiple entry points here. Users may start instead with `gddy domain suggest --interactive` and be presented with a search interface that then leads through the purchase flow.

docs/design/INTERACTIVE_DOMAIN_WIZARD.md
```diff
@@ -0,0 +1,483 @@
...
+  },
+  "system": "domain",
+  "next_actions": [
+    {"command": "domain get example.com", "description": "View domain details"},
+    {"command": "dns record add example.com ...", "description": "Configure DNS"}
+  ]
+}
+```
+
+This preserves scriptability: `gddy domain register 2>/dev/null` gives clean JSON even when the interactive wizard ran.
```
jpage-godaddy: Personally, I prefer the human output mode for interactive flows. Preserving scriptability seems moot to me if you're doing an interactive experience.

rts1-godaddy: Agreed. I will update the design to

- Interactive mode → human output by default (summary, colors, next-steps guidance on stdout)
- Non-interactive mode (--non-interactive or no TTY) → JSON envelope as usual for scripting/agent consumption
- User can still force JSON with --output json if they want structured output after an interactive session (edge case, but supported)

docs/design/INTERACTIVE_DOMAIN_WIZARD.md
rts1-godaddy: Good call, updated the design to support this. Any domain subcommand can hand off to the purchase wizard via `--interactive`:

* gddy domain suggest "cool startup" --interactive → pick from suggestions → continue to purchase
* gddy domain available example.com --interactive → if available, continue to options/purchase
* gddy domain quote example.com --interactive → jump straight to review & confirm

The command does its normal work first, then instead of exiting, it feeds the result into the wizard and continues from the appropriate step. Same state machine, different starting points.

Commands that don't naturally lead to a purchase (domain list, domain get, domain contacts init) ignore the flag.

jpage-godaddy (approved):

PR #190, opened 2026-08-07
feat(environments): read DevX Core URL from config (DEVX-793)

smukherjee-godaddy:
## Summary

- read `devx_core_url` from each active `environments.toml` entry
- preserve `<ENV>_DEVX_CORE_URL` and `DEVX_CORE_URL` as higher-precedence runtime overrides
- retain built-in DevX Core URLs for `prod` and `ote`, and cover file configuration, precedence, and invalid URLs

## Example

```toml
[dev]
api_url = "https://api.dev-godaddy.com"
client_id = "..."
devx_core_url = "https://api.developer.commerce.dev-godaddy.com"
```

## Validation

- `cargo check`
- `cargo clippy -- -D warnings`
- `cargo test` (552 passed)
- `cargo fmt --check`

https://godaddy-corp.atlassian.net/browse/DEVX-793

### DIFF
--- a/rust/src/environments/mod.rs
+++ b/rust/src/environments/mod.rs
@@ -18,6 +18,7 @@
 //! ```toml
 //! [dev]
 //! api_url = "https://api.dev-godaddy.com"
+//! devx_core_url = "https://api.developer.commerce.dev-godaddy.com"
 //! min_stage = "experimental"
 //!
 //! [staging.feature_overrides]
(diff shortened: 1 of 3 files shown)


rust/src/environments/mod.rs
```diff
@@ -172,6 +182,15 @@ fn default_account_url(sources: &SourceChain<'_>) -> String {
     derive_account_url(sources.env_name().unwrap_or_default())
 }
 
+fn default_devx_core_url(sources: &SourceChain<'_>) -> String {
```
jpage-godaddy: It'd be simpler if you added the config value to `BUILTIN_ENVS` and extended `BaseEnvConfig`; then you don't need a separate environment-to-values lookup table or this function. The default function is mainly for if you needed to compute a default; for pre-populating built-in environments, we already have a mechanism for that.

smukherjee-godaddy: Good call, thanks. I’ll move the built-in DevX Core URLs into `BUILTIN_ENVS` and remove the separate lookup/default function, while retaining the existing environment-variable override precedence.

jpage-godaddy (approved):

PR #192, opened 2026-08-07
chore(rust): enforce command-tree file structure; split oversized modules

jpage-godaddy:
## Summary
- Documents the desired file structure (one file per subcommand, mirroring the CLI command tree — see `domain`/`dns` as the reference shape) in `docs/code-structure.md`, with a pointer from `AGENTS.md`.
- Adds `rust/scripts/check-module-size.sh`, wired into CI, which fails the build if any hand-written `.rs` file exceeds 1000 lines. No allowlist.
- Splits every file that was already over that limit into per-subcommand/per-concern files: `dns/set.rs`, `hosting/nodejs/mod.rs`, `api_explorer/mod.rs`, `extension/mod.rs`, `application/commands/mod.rs`, and `tools/generate-api-catalog/src/main.rs`.

## Test plan
- [x] `cargo check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test` — 550 passed, same count as before the refactor across every touched module
- [x] `cargo fmt --check` — clean
- [x] `./rust/scripts/check-module-size.sh` — zero violations
- [x] Manually exercised `dns set`, `hosting nodejs app list`, `platform app list`, `api search`, `api call` against a live environment to confirm the refactored wiring/auth/client code behaves identically


docs/code-structure.md
```diff
@@ -0,0 +1,31 @@
+# Code file structure
+
+`gddy`'s Rust source should mirror its CLI command tree. This doc explains the target shape and the CI check that enforces it, with a worked example of what to do — and what not to do.
+
+## The rule
+
+- One file per leaf subcommand, named after the subcommand (`domain/purchase.rs` for `gddy domain purchase`).
+- One `mod.rs` per command group, containing *only* wiring: `mod` declarations, a `pub fn module()` (top level) or `pub(super) fn group()` (nested), and nothing that could stand alone as a unit of behavior.
+- Shared cross-cutting helpers (an authenticated client, error-body rendering, money/format helpers, validation) live in a sibling `common.rs`, exposed at `pub(super)`/`pub(crate)` — not duplicated per subcommand file, and not left in `mod.rs`. If the `common.rs` file grows too large, like over 1000 lines of code, split it in to appropriately named modules per theme.
```
runruh-godaddy: NIT: Should this be declarative of a limit and not a `like over 1000 lines of code` giving room for interpretation?

PR #193, opened 2026-08-07
docs: generalized interactivity design proposal

jpage-godaddy:
## Summary
- Generalizes the [Interactive Domain Purchase Wizard](docs/design/interactive_domain_wizard.md) proposal into a consistent pattern for interactivity across all `gddy` commands, triggered by missing required inputs rather than one-off wizards.
- Proposes `--interactive`/`--non-interactive` flags, defaulted by TTY detection the same way `--output` already is.
- Leaves library selection as an open question for discussion: inline/scrollback prompt libraries (inquire, dialoguer, promkit) vs. full-screen TUI (ratatui, cursive), with pros/cons of each and no recommendation locked in yet.

## Test plan
- [ ] N/A — documentation only, no code changes
- [ ] Design discussion / feedback from reviewers on the open library-selection question


rts1-godaddy:
I like the direction here, "interactivity triggered by missing inputs" is the right model and aligns with how `gh` and `Shopify` CLI handle this.

One thing I think would strengthen this doc is explicitly distinguishing between the two cases it covers, since they have very different implementation complexity:

1. **Simple auto-prompts:** A command is missing a required arg → prompt for it, get the value, execute. This is what most commands would use. It could live entirely in the engine (intercept clap's "missing arg" error, prompt, re-dispatch). Command authors get it for free.
2. **Multi-step flows:** Commands like `domain register` where prompts depend on the results of earlier API calls, state accumulates across steps, and there's conditional branching (domain taken → show suggestions). These need handler-level orchestration with a state machine.

Both fit under the "interactivity triggered by missing inputs" umbrella, but the first is an engine level primitive (nearly zero work for command authors) while the second requires explicit wizard code in the handler. Calling this out would help clarify what command authors get automatically vs. what they need to build themselves.

docs/design/generalized_interactivity_design.md
```diff
@@ -0,0 +1,83 @@
...
+    C -->|Yes| D[Execute the command]
+    C -->|No| E[Present an error]
+```
+
+A user that has all of the required inputs does not want or need an interactive experience; this is especially common in the case of writing a shell script to automate tasks. A user that isn't already familiar with the input requirements, however, most likely does not prefer receiving error messages with corrective guidance. Instead, we can prompt the user to fill in what is needed interactively so they don't have to re-invoke a corrected command. We can follow this algorithm instead:
+
+```mermaid
+flowchart TD
+    A[Parse the command that was invoked] --> B["Create collection `inputs` of supplied inputs"]
+    B --> C[For each required input, in user-friendly order]
```
rts1-godaddy, line 24: Looking at the existing commands, they are already defined in a logical sequence, but I don't see any docs around this. I am actively building a Comprehensive guide to add new commands, I will update that doc and I should be able to share that doc by tomorrow for further review.

docs/design/generalized_interactivity_design.md
```diff
@@ -0,0 +1,83 @@
...
+    D -->|Yes| E["Place the input in `inputs`"]
+    E --> C
+    D -->|No| F[Start an interactive terminal experience for collecting the data]
+    F --> G{Did the user complete the experience?}
+    G -->|Yes| H["Store the form data into `inputs`"]
+    H --> C
+    G -->|No| I[Present an error telling the user how to re-invoke the command with their prior inputs preserved, so they can resume later]
+```
+
+So essentially, interactivity is triggered by missing command inputs. Users can either do a one-shot command execution or be walked through a wizard experience.
```
rts1-godaddy, line 35: what happens for mutating commands when all inputs are present but we are in interactive mode?

like `gddy domain purchase --quote-token abc --agree --confirm`, all required inputs satisfied. Today `.mutates(true)` only enables `--dry-run` short-circuit, and `--confirm` is manually checked in the handler. There is no engine-level `are you sure?` prompt.

With the generalized interactivity model, this seems like a natural place for the engine to add a confirmation check: if a command is `.mutates(true)` and interactive mode is on and `--confirm` wasn't explicitly passed, auto-prompt `Confirm [action]? [y/N]` rather than requiring the handler to implement it manually. This would make confirmation consistent across all mutating commands without each one hand-rolling the logic.

jpage-godaddy: Love that idea

docs/design/generalized_interactivity_design.md
```diff
@@ -0,0 +1,96 @@
...
+
+**[ratatui](https://ratatui.rs)** (~42M downloads) is the dominant Rust TUI crate, but it's immediate-mode and low-level: you own the render loop and redraw the full frame yourself, with no built-in "pick from a list" or dialog widgets. Anything wizard-shaped (steps, forms, cart summary) would be built from primitives.
+
+**[cursive](https://docs.rs/cursive)** (~1.7M downloads) is also full-screen but retained-mode and higher-level: a `View` trait plus a tree of composed views, with dialogs, menus, buttons, layout managers, and theming shipped in, driven by callbacks. It supports four backends (crossterm, ncurses, termion, pancurses). For a forms/dialog-heavy wizard specifically, cursive gets us closer to off-the-shelf than ratatui does; ratatui's advantage is its larger ecosystem and lower-level control if cursive's opinions ever get in the way.
+
+### Open question
+
+This proposal doesn't yet commit to a side. The two live paths are:
+
+1. **Stay inline** with inquire/dialoguer for simple prompts, and reach for promkit if/when a step needs a genuinely custom widget — accepting promkit's smaller footprint as the tradeoff for keeping scrollback.
```
rts1-godaddy, line 82: +1 for Inline with scroll back preserved. We can probably include the Full-screen in future if we ever needed.

For suggestions and missing arguments, in-line works. 

For wizards, inline + a thin step-runner + colored section headers between prompts gets us 90% of the UX without alt-screen

rts1-godaddy (approved):

PR #200, opened 2026-08-11
feat(api-explorer): add `api graphql` command group for GraphQL-backed domains

jpage-godaddy:
## Summary
- Some API domains (e.g. `taxes`, `catalog-products`) are backed by GraphQL rather than plain REST.
- `api operation list --domain <domain>` and `api search` now tag each addressable query/mutation field with a `kind`.
- New `api graphql` command group gives GraphQL operations their own surface: `get`, `call`, `type`, and `sdl` subcommands — since REST-shaped commands (`api operation get`, `api call`) don't fit a GraphQL operation.
- `api operation get`/`api call` redirect to the new group when pointed at a GraphQL operation id.
- Regenerated `catalog-products` and `taxes` API catalog schemas to include GraphQL operation metadata.

### Refined during review
- `--select` on `api graphql call` is now validated as comma/dot-separated GraphQL field names before being concatenated into the synthesized query — an unvalidated value could otherwise inject arbitrary GraphQL syntax into the request.
- A list-typed `--arg` value that isn't valid JSON is now a validation error instead of silently being sent as a JSON string (which the server would reject with a confusing type error).
- `--header KEY:VALUE` now rejects an empty key (e.g. `:value`) up front with the CLI's own validation error, rather than letting reqwest fail later with a less-actionable message.
- Added test coverage for the shared HTTP status/GraphQL-error/403/non-2xx handling in `send_and_report`.

## Test plan
- [x] `cargo check`
- [x] `cargo test` (577 passed)
- [x] `cargo fmt --check`
- [x] Verified `--select` injection payloads and empty header keys are rejected with clear validation errors (manual + regression tests)
- [ ] Manual smoke test of `gddy api graphql get/call/type/sdl` against a live GraphQL-backed domain

### DIFF
--- a/rust/schemas/api/taxes.json
+++ b/rust/schemas/api/taxes.json
@@ -942,7 +942,2273 @@
             "returnType": "Rate"
           }
         ],
-        "schemaRef": "./graphql/schema.graphql"
+        "schemaRef": "./graphql/schema.graphql",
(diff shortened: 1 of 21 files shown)


rust/src/api_explorer/graphql/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+mod call;
+mod get;
+mod sdl;
+mod type_cmd;
+
+pub(super) fn group() -> RuntimeGroupSpec {
+    RuntimeGroupSpec::new(
+        GroupSpec::new("graphql", "Inspect and call individual GraphQL operations").with_long(
+            "Some API domains are backed by GraphQL rather than REST — \
+            `api operation list  --domain <domain>` and `api search` tag each \
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in f148781: removed the doubled space in the `--help` text on both lines.

rust/src/api_explorer/guides/api-explorer.md
```diff
@@ -104,15 +104,86 @@ gddy api call <path> --method POST --file body.json
 
 Before running anything that changes data, preview it with the global `--dry-run` flag — it short-circuits before sending the request for any method that mutates state (a plain `GET`/`HEAD` still runs for real under `--dry-run`, since there's nothing unsafe to preview there).
 
+## GraphQL operations
+
+Some domains are backed by GraphQL rather than plain REST. In the catalog they still show up as one wrapper operation (`postTaxGraphql`, `postCatalogGraphql`) that accepts a generic `{query, variables}` body — but every GraphQL query and mutation that wrapper proxies to is also individually addressable.
+
+A GraphQL operations have their own dedicated set of commands. Run this to show a GraphQL operation's details:
```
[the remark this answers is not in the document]
jpage-godaddy: Fixed in f148781: reworded to "GraphQL operations have their own..." and dropped a stray trailing space on the next paragraph.

rust/src/api_explorer/graphql/sdl.rs
```diff
@@ -0,0 +1,117 @@
...
+            `postTaxGraphql`), not an individual query/mutation id.",
+        )
+        .with_system("api")
+        .with_tier(Tier::Read)
+        .no_auth(true)
+        .raw_output(true),
+        |_cred, args: GraphqlSdlGetArgs| async move {
+            let operation = args.operation.as_str();
+            let (_, ep) = resolve_operation(catalog(), operation, None)?;
+            let schema = ep.graphql.as_ref().ok_or_else(|| {
```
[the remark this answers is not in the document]
jpage-godaddy: I tried to reproduce this and couldn't — `resolve_operation`'s fuzzy fallback (`search_endpoints`) checks `haystack.contains(&q)` where `q` is the full query. A compound GraphQL operation id like `postTaxGraphql::query::classification` is longer than any wrapper endpoint's operation_id/path/summary/description, so it can never be a substring match there, and it doesn't match the `kind name`-formatted GraphQL-ops haystack either (that string has no `::`). Confirmed live: `gddy api graphql sdl get postTaxGraphql::query::classification` returns a clean `NOT_FOUND` ("no operation found matching..."), not SDL for the wrong operation. `gddy api search` with the same id also returns zero hits. Leaving as-is since I can't reproduce a wrong match — happy to revisit if you have a concrete id that does misfire.

rust/src/api_explorer/graphql/call.rs
```diff
@@ -0,0 +1,559 @@
...
+            .filter(|a| variables.contains_key(&a.name))
+            .map(|a| format!("${}: {}", a.name, a.arg_type))
+            .collect();
+    let call_args: Vec<String> =
+        g.op.args
+            .iter()
+            .filter(|a| variables.contains_key(&a.name))
+            .map(|a| format!("{0}: ${0}", a.name))
+            .collect();
+    let selection = build_selection(args.select.as_deref(), &g.op.return_type);
```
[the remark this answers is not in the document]
jpage-godaddy: Good catch — fixed in fccc4e6. Added `validate_selection_expr` which rejects any `--select` value whose comma/dot-separated segments aren't syntactically valid GraphQL field names, before it ever reaches the query-string concatenation. Added a regression test with a braces/mutation-injection payload confirming it's now a validation error instead of being embedded in the query.

rust/src/api_explorer/guides/api-explorer.md
```diff
@@ -104,15 +104,86 @@ gddy api call <path> --method POST --file body.json
...
+
+If you'd rather have a structured, JSON-native dump instead of the raw SDL text — every operation and every type, with full field/arg detail and descriptions, already broken into our summary shape — that's available too:
+
+```
+gddy api operation get postTaxGraphql --output json
+```
+
+The result's `graphql` field has it all: `operations` and `types`. This is the same data `api graphql get`/`api graphql type get` read from, just all at once instead of one hop at a time — useful if you want to script against the JSON shape rather than parse GraphQL SDL yourself.
+
+The wrapper operation itself (`postTaxGraphql`, `postCatalogGraphql`) keeps working exactly as before under the regular REST commands, if you'd rather hand-write the GraphQL query text yourself — `gddy api call postTaxGraphql --body '{"query": "...", "variables": {...}}'` is unaffected by any of the above.
```
rts1-godaddy: when i run this command `cargo run -- api call postTaxGraphql --body '{"query": "{ something }"}'`, I am getting following error 

> Error: endpoint must be a URL path starting with '/', or a GraphQL operation id (see `api operation list --domain <domain>`), not "postTaxGraphql" — use `api operation get postTaxGraphql` to findthe concrete path or id

I think the `resolve_graphql_operation` -> `parse_graphql_op_id` expecting it to be `postTaxGraphql::query::classification` and since we are only providing `postTaxGraphql` it fails ?

jpage-godaddy: Good catch, thanks! Your diagnosis was half right, but the actual bug is one level up from `parse_graphql_op_id`: `api call` has always required a *concrete path* (starting with `/`), never a bare operationId — that's pre-existing validation, not something this PR touched. `postTaxGraphql` is the wrapper's REST operationId, not a compound GraphQL id (`postTaxGraphql::query::classification`), so `resolve_graphql_operation`/`parse_graphql_op_id` correctly returns None for it and the redirect never fires — you just land on the generic 'must be a URL path' error instead.

The real problem was my guide example: it claimed `gddy api call postTaxGraphql --body '...'` 'keeps working exactly as before,' which was never true — `api call` needs the wrapper's actual `fullPath` (e.g. `/v2/commerce/stores/{storeId}/tax-subgraph`, from `api operation get postTaxGraphql`), not its operationId. Fixed in 9f6d8c5 to show the concrete path instead. Verified the corrected form passes validation under `--dry-run`.

jpage-godaddy: Created DEVEX-1008 because we _ought to_ be able to call REST APIs with an operation ID.

rust/schemas/api/taxes.json
```diff
@@ -942,7 +942,2273 @@
...
+              {
+                "name": "updatedAt",
+                "type": "OrderByDirectionEnum"
+              }
+            ],
+            "kind": "input",
+            "name": "ReferenceOrderByInput"
+          },
+          {
+            "fields": [
```
rts1-godaddy, line 3131: This is conflicting with an existing `ReferenceValueFilter` that has an extra field. When i run `gddy api graphql type get ReferenceValueFilter`, its returning the schema from `catalog-products` .

jpage-godaddy: Gah. I think for REST we ensure schemas are uniquely identified; need to fix that for GraphQL types too. Good catch.

jpage-godaddy: Good catch — confirmed, this is a real bug, not a coincidence. `taxes` and `catalog-products` each declare their own `ReferenceValueFilter` input type (taxes' has only `in`; catalog-products' has `eq` and `in`), and `find_graphql_type` searched the whole catalog flattened into one namespace, silently returning whichever domain sorted first alphabetically. GraphQL type names are only unique within a single subgraph, so this could misfire for any other name collision too.

Fixed in 59a8f1b: `api graphql type get` now collects every domain that defines the name, and without `--domain` to disambiguate it errors out listing all of them (`AMBIGUOUS_MATCH`) instead of guessing. `--domain taxes`/`--domain catalog-products` each now return that domain's own fields. Added a regression test pinned to this exact `ReferenceValueFilter` collision.

rust/src/api_explorer/graphql/call.rs
```diff
@@ -0,0 +1,609 @@
...
+            if ctx.dry_run() && g.op.kind == "mutation" {
+                return Ok(CommandResult::new(json!({
+                    "command": "api:graphql:call",
+                    "action": "dry-run: would execute",
+                    "operation": id,
+                }))
+                .with_dry_run());
+            }
+
+            let built = build_graphql_call(&g, &args)?;
```
qcai-godaddy: It might still be worth running `build_graphql_call` first so missing/invalid `--arg/--select` fail under `--dry-run` the same way they would on a real call, where the request is validated even if nothing is applied.

rust/src/api_explorer/guides/api-explorer.md
```diff
@@ -104,15 +104,88 @@ gddy api call <path> --method POST --file body.json
...
+```
+gddy api graphql call postTaxGraphql::query::classification --arg storeId=<uuid> --arg x-store-id=<uuid> --arg id=<id>
+```
+
+The CLI builds the GraphQL query text and `variables` object for you, sends it to the HTTP wrapper, and reports the response or errors.
+
+By default the response selects just `{ __typename }` for an object-shaped return type (or nothing at all for a plain scalar). To ask for specific fields, use `--select`, comma-separated and dot-nested for fields on a nested object:
+
+```
+gddy api graphql call postTaxGraphql::query::classification --arg id=<id> --select id,name,rate.percentage
```
qcai-godaddy: Should this example include the required `storeId` and `x-store-id` values from the example above? copying and pasting it will fail validation.

rust/src/api_explorer/graphql/call.rs
jpage-godaddy: Good call — fixed in 8e90918. `build_graphql_call` now runs before the dry-run short-circuit, matching how `api call`'s own method/endpoint checks already validate unconditionally under `--dry-run`. Added a regression test asserting a dry-run mutation with missing required `--arg` now fails with the same validation error a real call would, instead of a successful 'would execute' preview.

rust/src/api_explorer/guides/api-explorer.md
jpage-godaddy: Yep, good catch — fixed in 8e90918, added the storeId/x-store-id args so it's copy-pasteable.

qcai-godaddy (approved):

PR #206, opened 2026-08-12
fix(platform): list apps across all App Registry statuses

dsingh5-godaddy:
## Summary
- `gddy platform app list` now sends an explicit App Registry `status.in` for every `ApplicationStatus` (`ACTIVE`, `ARCHIVED`, `BLOCKED`, `INACTIVE`, `VERIFYING`).
- Fixes ACTIVE-only results caused by the API defaulting when `status` is omitted.

## Test plan
- [ ] `cargo test --lib application::commands::list application::client::tests::list_applications`
- [ ] `cargo run -- platform app list --env dev` returns non-ACTIVE apps when present
- [ ] Confirm help text mentions listing across every App Registry status

Jira: [DEVX-837](https://godaddy-corp.atlassian.net/browse/DEVX-837)

### DIFF
--- a/rust/src/application/commands/list.rs
+++ b/rust/src/application/commands/list.rs
@@ -12,8 +12,9 @@ pub(super) fn command() -> RuntimeCommandSpec {
         CommandSpec::new("list", "List all applications")
             .with_long(
                 "List all GoDaddy developer-platform applications visible to \
-                the current account. Use `gddy platform app info --name <name>` \
-                to fetch full details for a single application.",
+                the current account, across every App Registry status. Use \
+                `gddy platform app info --name <name>` to fetch full details \
+                for a single application.",
             )
             .with_system("applications")
             .with_tier(Tier::Read)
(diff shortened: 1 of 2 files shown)


grajput-godaddy (approved):

rust/src/application/commands/list.rs
```diff
@@ -12,8 +12,9 @@ pub(super) fn command() -> RuntimeCommandSpec {
         CommandSpec::new("list", "List all applications")
             .with_long(
                 "List all GoDaddy developer-platform applications visible to \
-                the current account. Use `gddy platform app info --name <name>` \
-                to fetch full details for a single application.",
+                the current account, across every App Registry status. Use \
```
jpage-godaddy, line 15: Wouldn't "all" already imply it's across all statuses?

jpage-godaddy (approved):

rust/src/application/commands/list.rs
dsingh5-godaddy: "all" isn’t a valid status — the enum is only ACTIVE | ARCHIVED | BLOCKED | INACTIVE | VERIFYING. 

Omitting status also isn’t “all”: the API defaults to ACTIVE-only. 

The CLI has to send status.in with every enum value, which is why the help text calls that out.

PR #212, opened 2026-08-13
feat(email): Added CLI support for panel based email services

sgimpel-godaddy:


docs/proposals/email-management-cli.md
```diff
@@ -0,0 +1,176 @@
...
+
+When the response is ineligible, or an eligible account carries outstanding
+`requirements`, attach a `next_actions` entry pointing at `email create` with
+the relevant `--account-id`/`--consent` pre-filled from the response.
+
+### `gddy email create`
+
+| | |
+|---|---|
+| Flags | `--email <email>` (required), `--account-id`, `--first-name`, `--last-name`, repeatable `--consent <agreementType>` |
```
jpage-godaddy: What is account ID?

docs/proposals/email-management-cli.md
```diff
@@ -0,0 +1,176 @@
...
+### 3. List pagination UX: native `--page`/`--page-size` vs generic `--limit`/`--offset`
+
+The server paginates natively (`page`/`pageSize`, capped at 100, with
+HATEOAS `links`), which doesn't fit cli-engine's `PaginationConfig` model —
+that model expects the handler to return the *complete* list and lets the
+engine slice it via generic `--limit`/`--offset`, the way `domain list` does.
+
+**Recommendation: native `--page`/`--page-size` flags, forwarded 1:1 to the
+API's query params**, rather than adopting `--limit`/`--offset` via
+`PaginationConfig`. Translating page-based server pagination into offset math
```
jpage-godaddy: I'm pretty hesitant to have mismatching pagination styles. I think that `--limit`  maps pretty cleanly to `--page-size`, and if `--offset` is a multiple of `--page-size`, then it should typically work the same (worst case scenario is that you have to fetch two pages and then slice them up to fit the provided offset/limit). We can specify default limits now, so the assertion that the handler must return the entire list isn't accurate (maybe that was based on an older version of the engine since enhanced pagination functionality is now available). We _should_ be able to, for this `list` command have a default page size/limit, and if we wanted to add additional constraints around `--limit` and `--offset` (like forcing them to be page-aligned), that'd be better than having completely different pagination flags.

Worth discussing with the others.

rust/src/email/check_eligibility.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    if let Some(account_id) = first_eligible_account_id(data) {
+        create = create.with_param("account-id", NextActionParam::value(account_id));
+    }
+
+    vec![create]
+}
+
+fn first_eligible_account_id(data: &Value) -> Option<String> {
+    data.get("eligibleAccounts")?
```
jpage-godaddy, line 59: Not blocking, but I think a good backlog enhancement would be to switch this client to being spec-generated so that we have strongly-typed structures and avoid accidental data shape mismatches.

rust/src/email/create.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::email::client::ClientError;
+use crate::email::{client_err, client_err_with_fix, make_client};
+use crate::scopes::EMAIL_CREATE;
+
+#[derive(Debug, Clone, clap::Args)]
+struct CreateArgs {
+    #[arg(long, value_name = "EMAIL")]
+    email: String,
+    #[arg(long = "account-id", value_name = "ACCOUNT_ID")]
```
jpage-godaddy, line 16: We should have the `///` comments over all these arg so we can have better help text. I think `--account-id` especially needs an explanation.

jpage-godaddy:
I'd recommend attaching a markdown guide to the command module as well, just to explain the email system, things like what is an account ID, etc.

docs/proposals/email-management-cli.md
sgimpel-godaddy: It's a wopr based id, not a customerId or shopperId. Usually corresponds to some subscription information, orionUid, etc.

docs/proposals/email-management-cli.md
sgimpel-godaddy: Set back to `--limit`

rust/src/email/create.rs
sgimpel-godaddy: Added

rust/src/email/guides/email-mailboxes.md
jpage-godaddy, line 1: If you reformat this to do soft line breaks instead of hard breaks, then `cli-engine`'s markdown-to-terminal formatting takes care of word wrapping based on the terminal width instead.

rust/src/email/guides/email-mailboxes.md
```diff
@@ -0,0 +1,68 @@
+---
+summary: How GoDaddy Email mailboxes, accounts, and consent fit together
+---
+
+# GoDaddy Email mailboxes
+
+`gddy email` manages GoDaddy Email mailboxes over panel-v3, the newer
+OAuth-bearer-token API. It's a different system from domain or hosting
+"accounts" — this guide covers the vocabulary specific to email.
```
jpage-godaddy: This is leaking context relating to the GoDaddy work assignment; it should be targeted toward our end-users.


```suggestion
This guide explains the GoDaddy email system and how to use the `gddy email` commands to manage it.
```

rust/src/email/guides/email-mailboxes.md
```diff
@@ -0,0 +1,68 @@
...
+
+## What an "account" is here
+
+An **account** (`accountId`) identifies an existing GoDaddy Email/productivity
+account the customer already holds under panel-v3. It's an opaque ID scoped to
+this API — it is **not** a shopper/customer ID, and it has nothing to do with
+domain or hosting "accounts" elsewhere in `gddy`. A customer may hold zero,
+one, or several eligible email accounts; `email create` needs to know which
+one to provision the new mailbox under.
```
jpage-godaddy: Seems to be explaining things to us, not to users ("a customer" versus "you"). Is "productivity" a term that customers will understand? Will they understand "panel-v3?" Also, I'm not sure this explains an account sufficiently. Is there an account per domain name? Per inbox? This doesn't make it clear why we have accounts as a distinct thing from their login account or their domain name. I'm just wondering if account ID is something this CLI needs to have at all.

jpage-godaddy (approved):
Looks good once rebasing on main to resolve the merge conflict and 1000-lines issue for `environments/mod.rs`. You'll also want to fix the PR title to fit the conventional commit format (label commits/PR with a "feat(email):" prefix)

PR #216, opened 2026-08-18
feat(app): add settings support for app registration

nmolham-godaddy:
## Summary

- Add `[[settings]]` support to `godaddy.toml` and wire it through `gddy platform app add settings` / `gddy platform app release` — closes the gap where `app-registry-api`'s `createRelease.settings` capability had no CLI support at all (config, add command, and release payload all silently dropped it).
- New `gddy platform app add settings` command writes placement metadata only (`group`/`slug`/`entryPath`/`order`/`capabilities`/`icon`) — matches the flag shape of `add action`/`add subscription`/`add extension`.
- Full typed `settings-form-v1` presentation model (`text`/`textarea`/`number`/`boolean`/`select`/`multi-select`/`list-group`, recursive nesting) authored inline in `godaddy.toml` — no JSON file, no opaque passthrough. `presentation` is hand-added after `add settings`; `release` rejects a settings entry with no `presentation` instead of `Config::validate()`, so a placement-only entry still works fine for every other command.
- Structural client-side validation only (slug/entryPath shape, capability/icon-library enum membership, entry-path overlap, field-key uniqueness/pattern, non-empty `options`) — bounds/default-consistency/nesting-depth stay server-validated, consistent with how the rest of `Config::validate()` already treats every other section.
- New `rust/scripts/build-and-verify.sh` — check/clippy/fmt/module-size, then build and run `gddy` (args forwarded, defaults to `--help`).
- New `docs/application-settings.md` — end-user guide: workflow, full field-type reference, worked nested `list-group` example, gotchas (no release inheritance, no auto-backfill to already-enabled stores).

**Affects**: https://github.com/gdcorp-commerce/app-registry-api/pull/131

## Changes

- `rust/src/config/settings_form.rs` (new) — `SettingsFormV1Presentation`/`Section`/`Field` enum/`ListGroupItem`/`ChoiceOption`/`SelectValue`, structural validation.
- `rust/src/config/settings.rs` (new) — `SettingConfig`/`SettingIcon`, slug/entryPath/capability/icon-library validation, entry-path-overlap check.
- `rust/src/config/mod.rs` — `settings: Vec<SettingConfig>` field, wired into `Config::validate()`.
- `rust/src/application/commands/release.rs` — `setting_entry`/`build_settings` map config to `ApplicationSettingCreateInput`, enforce `presentation` required here (not in `Config::validate()`).
- `rust/src/application/commands/add.rs` — `add settings` command.
- `rust/src/application/commands/schemas.rs` — `ConfigSetting` output schema.
- `rust/src/application/commands/init.rs` — add missing `settings: vec![]` to a `Config` literal.
- `rust/scripts/build-and-verify.sh` (new).
- `docs/application-settings.md` (new).

## QA

- [x] `cargo check`
- [x] `cargo test` (19 new tests: config validation, `settings_form` structural checks + TOML round-trip, `release::setting_entry` mapping incl. missing-presentation error path)
- [x] `cargo fmt --check`
- [x] Manual: built binary, ran `add settings` against a scratch `godaddy.toml`, hand-added `[settings.presentation]` (incl. nested `list-group` with multiple `options`), confirmed round-trip through `read_config`/`validate`/`write_config` preserves `defaultValue` and every nested field; confirmed `--icon-name`/`--icon-library` pairing and entry-path-overlap rejections fire before any write.

## After merge
- [x] Update `docs/GPA-SETTINGS-REGISTRATION.md` in https://github.com/gdcorp-commerce/app-registry-api/pull/131 if applicable


smukherjee-godaddy:
@nmolham-godaddy thanks for driving this end-to-end — I validated the implementation against the live App Registry settings schema, the #131 registration plan, and DEVEX-1021.

I found two blockers before this can merge:

1. `cargo clippy -- -D warnings` currently fails on `clippy::collapsible_match` in `rust/src/config/settings_form.rs` (the select/multi-select options match arm).

2. `platform app release` treats *any* manifest read/parse/validation failure as non-fatal and sends empty `actions`/`subscriptions`/`uiExtensions`/`settings` arrays. With hand-authored settings presentations, a TOML typo or validation error could silently create a release without settings. Please preserve the empty fallback only for a missing manifest; parse and validation errors should fail the release.

Follow-up: the docs say local validation enforces key uniqueness, but nested `list-group` fields are not checked for duplicate keys, `idField` collisions, or an invalid `titleField` target. The App Registry rejects these, so either add the checks or narrow the documentation claim.

The core settings mapping and targeted GraphQL test otherwise align with App Registry.

wcole1-godaddy:
The API mapping here looks right, but we should support both presentation authoring modes before merging: inline TOML and a referenced JSON file.

Suggested manifest shape:

```toml
[[settings]]
group = "tax-center"
slug = "manual-tax"
entryPath = "/settings/manual-tax"
presentationFile = "fixtures/manual-tax-registry-presentation.json"
```

The referenced JSON should contain the complete API presentation object (`type`, `schemaVersion`, and `sections`), which lets existing GPA presentation fixtures be reused directly. Inline TOML can remain the current shorthand, with the CLI injecting `type = "form"` and `schemaVersion = "settings-form-v1"`.

Expected behavior:

- `presentation` and `presentationFile` are mutually exclusive.
- Either form is acceptable, but one is required at release time.
- Relative file paths resolve from the directory containing the selected `godaddy.toml`, not the caller's working directory.
- Missing, unreadable, malformed, or invalid JSON stops the release.
- Both forms go through the same typed validation and produce identical `createRelease.settings[].presentation` payloads.
- `gddy platform app add settings` should accept an optional `--presentation-file <path>`.
- Config rewrites preserve the file reference rather than embedding its contents.

This is especially useful for the existing GPA examples, which already keep substantial registry presentations in JSON fixtures; requiring inline-only TOML would create a second source of truth.

docs/application-settings.md
jpage-godaddy, line 1: It'd be better if this was embedded in a guide so that LLMs can discover these instructions without needing to do a web search. I think the whole `gddy platform` module probably needs a holistic guide with application settings included (or I guess multiple guides is also fine).

docs/application-settings.md
```diff
@@ -0,0 +1,149 @@
...
+    ```toml
+    [[settings.presentation.sections.fields.options]]
+    label = "United States"
+    value = "US"
+
+    [[settings.presentation.sections.fields.options]]
+    label = "Canada"
+    value = "CA"
+    ```
+  - Optional: `defaultValue` must match one option's `value`.
```
jpage-godaddy, line 81: Looks like the Markdown parser hit a snag starting on this line. You probably need to fix indentation or some other kind of white space.

rust/src/application/commands/add.rs
```diff
@@ -122,5 +162,67 @@ pub(super) fn group() -> RuntimeGroupSpec {
...
+    >(
+        CommandSpec::from_args::<SettingsArgs>(
+            "settings",
+            "Add an application settings placement to godaddy.toml",
+        )
+        .with_long(
+            "Register the placement metadata for an application-settings \
+            capability in the godaddy.toml manifest in the current directory. \
+            This command only writes group/slug/entryPath/order/capabilities/icon \
+            — it cannot author the settings-form-v1 form itself. After running \
```
jpage-godaddy, line 184: Is the meaning of "settings-form-v1" something that's clear to users?

rust/src/application/commands/release.rs
```diff
@@ -122,20 +172,22 @@ pub(super) fn command() -> RuntimeCommandSpec {
...
-                    (actions, subscriptions, ui_extensions)
+                    let settings = build_settings(&config)?;
+                    (actions, subscriptions, ui_extensions, settings)
                 }
                 Err(e) => {
                     tracing::warn!(
                         error = %e,
                         path = %config_path.display(),
-                        "failed to read config; releasing with empty actions, subscriptions, and uiExtensions"
+                        "failed to read config; releasing with empty actions, subscriptions, uiExtensions, and settings"
```
jpage-godaddy: I know this isn't new behavior, but if someone doesn't notice this warning, it could cause a lot of frustration, if they're releasing but it doesn't reflect their config. As a user, I'd prefer the release failed instead.

(non-blocking)

docs/application-settings.md
nmolham-godaddy: I agree, sounds like a good idea, I will do my best effort to implement it

nmolham-godaddy: [b298b7a](https://github.com/godaddy/cli/pull/216/commits/b298b7a6ad0a7c63261c532d1beb19360bd92d52) what do you think? CC @wcole1-godaddy

docs/application-settings.md
nmolham-godaddy: I can see rendering fine 🤔 what do you see when you open the file ?
https://github.com/nmolham-godaddy/gddy-cli/blob/2f77e886a0e5eb551747bd645d4a4b962cc03cbd/docs/application-settings.md#presentation-shape

Screenshot 2026-08-19 at 12 14 05 pm

rust/src/application/commands/add.rs
nmolham-godaddy: initially yes, but according to Wes's comment here, I should also add presentation authoring modes in this PR, so this description will change

rust/src/application/commands/release.rs
nmolham-godaddy: Agreed, fixed [c4c0115](https://github.com/godaddy/cli/pull/216/commits/c4c0115b06aecfe8d78fb10b2e9c994d864317ca) what do you think?

docs/application-settings.md
jpage-godaddy: I mean in the GitHub UI diff view. But if it renders fine, that's ok.

rust/src/application/commands/config.rs
```diff
@@ -0,0 +1,55 @@
...
+    ))
+    .with_command(RuntimeCommandSpec::new_with_context(
+        CommandSpec::new("validate", "Validate the local godaddy.toml manifest")
+            .with_long(
+                "Read the godaddy.toml (or godaddy.<env>.toml, per --env) \
+                manifest in the current directory and validate it against the \
+                same rules enforced when writing it via `platform app add`/ \
+                `init`/`release` — required fields, URL/UUID/semver shapes, \
+                and settings placement rules (slug/entry-path/capability \
+                enum). Reports every violation found, not just the first.",
```
jpage-godaddy: Man, I hate how Claude writes. It always highlights things you asked it to do recently instead of writing for the audience. If we list the specific things that are validated we'll just have to constantly update it. Seems like this can be simplified.

```suggestion
                "Validate the godaddy.toml (or godaddy.<env>.toml, per --env) \
                manifest in the current directory.",
```

rust/src/application/commands/release.rs
```diff
@@ -29,6 +34,129 @@ fn ui_extension_entry(
...
 }
 
+/// Resolves a setting's presentation from `presentation` or `presentationFile`.
+fn resolve_presentation(
+    setting: &crate::config::SettingConfig,
+    manifest_dir: &Path,
+) -> cli_engine::Result<SettingsFormV1Presentation> {
+    match (&setting.presentation, &setting.presentation_file) {
+        (Some(_), Some(_)) => Err(crate::error::GddyError::validation(format!(
+            "settings '{}' has both presentation and presentationFile — provide only one",
```
jpage-godaddy:
```suggestion
            "setting '{}' has both presentation and presentationFile — provide only one",
```

rust/src/application/commands/release.rs
```diff
@@ -29,6 +34,129 @@ fn ui_extension_entry(
...
+            setting.slug
+        ))
+        .into_cli_error()),
+        (Some(p), None) => Ok(p.clone()),
+        (None, Some(file)) => {
+            let path = manifest_dir.join(file);
+            let content = std::fs::read_to_string(&path).map_err(|e| {
+                crate::error::GddyError::validation(format!(
+                    "settings '{}' presentationFile {} could not be read: {e}",
```
jpage-godaddy: This is for a singular setting, no?

```suggestion
                    "setting '{}' presentationFile {} could not be read: {e}",
```

rust/src/application/commands/release.rs
```diff
@@ -29,6 +34,129 @@ fn ui_extension_entry(
...
+                    setting.slug,
+                    path.display()
+                ))
+                .into_cli_error()
+            })?;
+            presentation_from_json(&content).map_err(|e| {
+                crate::error::GddyError::validation(format!(
+                    "settings '{}' presentationFile {} is invalid: {e}",
```
jpage-godaddy:
```suggestion
                    "setting '{}' presentationFile {} is invalid: {e}",
```

rust/src/application/commands/release.rs
```diff
@@ -29,6 +34,129 @@ fn ui_extension_entry(
...
+                    setting.slug,
+                    path.display()
+                ))
+                .into_cli_error()
+            })
+        }
+        (None, None) => Err(crate::error::GddyError::validation(format!(
+            "settings '{}' has no presentation — add a [settings.presentation] block or a presentationFile before releasing",
```
jpage-godaddy:
```suggestion
            "setting '{}' has no presentation — add a [settings.presentation] block or a presentationFile before releasing",
```

rust/src/application/commands/release.rs
```diff
@@ -29,6 +34,129 @@ fn ui_extension_entry(
...
+fn setting_entry(
+    setting: &crate::config::SettingConfig,
+    manifest_dir: &Path,
+) -> cli_engine::Result<Value> {
+    let presentation = resolve_presentation(setting, manifest_dir)?;
+    let mut errors = Vec::new();
+    validate_presentation(&presentation, &mut errors, "presentation");
+    if !errors.is_empty() {
+        return Err(crate::error::GddyError::validation(format!(
+            "settings '{}' presentation is invalid: {}",
```
jpage-godaddy:
```suggestion
            "setting '{}' presentation is invalid: {}",
```

rust/src/platform/guides/platform-overview.md
jpage-godaddy, line 1: Well done on the guide

jpage-godaddy (commented):
Short of help text tweaks, the rest looks good to me.

rust/src/application/commands/config.rs
nmolham-godaddy: I agree 100%

nmolham-godaddy:
@jpage-godaddy all requested changes implemented

jpage-godaddy (approved):

PR #222, opened 2026-08-21
docs: Proposal for client resiliency

lmartinez1-godaddy:
Adds `docs/design/client_resiliency.md`, a proposal for owning retry, timeout, and rate-limiting policy in the CLI's HTTP layer instead of leaving it to callers.

No code changes. This is a policy doc to argue about before any of it gets built.

### DIFF
--- a/docs/design/client_resiliency.md
+++ b/docs/design/client_resiliency.md
@@ -1,0 +1,192 @@
+# Client Resiliency in the GoDaddy CLI
+
+How much human review this doc has had:
+
+- Origin: Human-Written / **AI-Assisted**
+- Review: Unread / Skimmed / **Read in Full**
+- Accuracy: Unverified / **Believed Accurate** / Fully Verified
+- Content: Untrimmed / **Trimmed**
+
+We are building an agent-first CLI and shipping it to customers as the primary programmatic surface for the GoDaddy Developer Platform.
+Every resiliency decision we make in `gddy` becomes load characteristics on our own APIs, multiplied by every customer who installs it.
+
+This proposal argues that retry behavior belongs to the library rather than the caller, states the risk we carry today, and lays out an ordered set of changes across `gddy` and `cli-engine`.
+
+## The risk
+
+Client-side retries are multiplicative, invisible to the caller, and worst exactly when the system is already degraded.
+GitHub's [August 17 outage](https://github.blog/news-insights/company-news/the-august-17-outage-and-the-work-ahead/) is a current example: a client-side retry loop against Copilot services amplified traffic during an already-degraded window and extended an outage to nearly eight hours.
+Their stated remediation is consistent retry limits, retry budgets, and variable timeouts enforced across service-to-service calls, not left to each caller.
+
+We have three distinct exposures.
+
+### 1. Unbounded hangs
+
+Three of our four HTTP client constructors set a user agent and nothing else.
+
+- `make_http_client()` at `rust/src/application/client.rs:16`, used by `application`, `hosting`, `webhook`, `api_explorer`, and `graphql`
+- `client_with_auth()` at `rust/domains-client/src/lib.rs:135`, used by every `domain` and `dns` command
+- `OnboardingClient::new()` at `rust/src/onboarding/client.rs:28`
+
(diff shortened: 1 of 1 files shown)


docs/design/client_resiliency.md
```diff
@@ -0,0 +1,192 @@
...
+An in-process token bucket on the AWS standard constants.
+
+- 500 token capacity, starting full
+- 14 tokens per transient retry, 5 per throttling retry
+- Refund the consumed cost when a retry succeeds, plus 1 token on a first-try success
+- When the budget empties, return the error without retrying
+
+This turns "retry 3 times" from an unconditional behavior into one that yields during a broad failure.
+With 3 max attempts it starts draining at roughly 22% sustained transient failure, so normal operation never notices it.
+It is the item aimed most directly at the failure mode in the GitHub postmortem.
```
jpage-godaddy, line 142: I'm not clear on the scope of this. The CLI is typically one or two requests per command, and then it's done. It's not a long-running service making long sequences of requests. If the concern is we're retrying too frequently, wouldn't 500 tokens with a cost of 5 or 14 per retry be far more than the current retry rate we're doing? Is this a persisted token bucket, lasting between CLI invokes? Or are you anticipating that when we add default request timeouts that we'll end up doing larger retry loops than normal so we need a 500 token budget to cover that?

jpage-godaddy: (ok I see the persist note later on. I like that idea)

docs/design/client_resiliency.md
```diff
@@ -0,0 +1,192 @@
...
+
+### 5. Honor `Retry-After`, in-process and across invocations
+
+When a 429 or 503 carries `Retry-After`, use the server's value clamped the way AWS clamps it: no lower than the computed backoff, no higher than computed backoff plus 5s, and no jitter applied on top since the server is expected to jitter it.
+
+Also persist it.
+Write a single "do not call `{host}` before `T`" timestamp per environment under `~/.config/gddy/`.
```
jpage-godaddy, line 149: Should we make this per host? One thing that makes it tricky is that everything goes through api.godaddy.com, so it's essentially one host for everything. I suppose of EFD itself is down that makes sense, but if the concern is a single system, we may want to scope to specific paths.

docs/design/client_resiliency.md
```diff
@@ -0,0 +1,192 @@
...
+- Refund the consumed cost when a retry succeeds, plus 1 token on a first-try success
+- When the budget empties, return the error without retrying
+
+This turns "retry 3 times" from an unconditional behavior into one that yields during a broad failure.
+With 3 max attempts it starts draining at roughly 22% sustained transient failure, so normal operation never notices it.
+It is the item aimed most directly at the failure mode in the GitHub postmortem.
+
+### 5. Honor `Retry-After`, in-process and across invocations
+
+When a 429 or 503 carries `Retry-After`, use the server's value clamped the way AWS clamps it: no lower than the computed backoff, no higher than computed backoff plus 5s, and no jitter applied on top since the server is expected to jitter it.
```
jpage-godaddy, line 146: I would throw 504s in here as well; I think we should treat them like 429s since they usually also reflect overloaded systems that are too exhausted to do a proper 429 response or that aren't configured with load shedding.

jpage-godaddy:
I like this idea. I had hoped that we could somehow easily use a shared HTTP "agent" across all API clients; that would allow us to do declarative sharing of HTTP behaviors. The trickiest bit is the spec-generated clients. There was a half-baked thing in place to set the user agent globally, but I'd love it if the generated clients could somehow share the code.

docs/design/client_resiliency.md
lmartinez1-godaddy: Good call...I don't know enough about api.godaddy.com and how the routes are configured. Is there some order to the way services underneath are segmented? If so, we could match that logic in the CLI to bifurcate more smoothly. 

If we have some signal that the issue truly is in api.godaddy.com, we could specifically return a signal from that front door service and treat it as a global problem.

jpage-godaddy:
Just going to merge this to get the PR out of the queue.

PR #248, opened 2026-09-04
feat: Add CLI command groups for Hosting API

jhenderson1-godaddy:
## Summary

Adds a new `hosting` command group running off the GoDaddy Hosting API (`/v1/hosting`). This sits alongside the existing (deprecated) `hosting nodejs` commands, which target the `/v1/hosting/nodejs` API (now deprecated).

**New commands (47 files, ~3500 lines):**
- `hosting app` – list, get, create, update, delete, restart, status
- `hosting deployment` – publish, get, list
- `hosting source` – upload (ZIP), github (import from branch), status
- `hosting subscription` – list, get, attach app
- `hosting domain` – list, get, attach to app, detach
- `hosting secrets` – list, create, update, delete, sync
- `hosting log list`
- `hosting runtime get`
- `hosting operation get`

All commands include `with_output_schema` and `with_default_fields` for table rendering.

## Design decisions

**One hosting module.** `hosting/nodejs/` (which targeted `/v1/hosting/nodejs`) has been removed to make way for the new `hosting/` module, which targets the new and generic `/v1/hosting` API.

**Bare arrays from list commands.** List commands collect all pages into a `Vec<Value>` and return `json!(all_items)` directly. Wrapping in `{ "items": [...] }` causes cli-engine to treat the response as a single row rather than iterating it as a table.

## Testing and Validation

Manual E2E testing covered the full lifecycle against the production environment:

1. Create Node.js app
2. Upload ZIP source and poll for completion
3. Import from GitHub branch on a git-connected app
4. Publish app and poll until complete
5. List hosting subscriptions (plans) and attach an app to one
6. Fetch preview and publish logs
7. Attach a custom domain (managed within GoDaddy)
8. Create, sync, list, and delete secrets
9. Check runtime status

All tests passed on prod (`GDDY_MIN_STAGE=beta`). Further follow-up tests were run, with the following commands verified: `hosting app list`, `hosting domain list`, `hosting log list`.

Automated testing covers the following:

Client tests (client_tests.rs) – 28 tests
  One test per HTTP method in the client, using a mock HTTP server. They verify path, query params, request body shape, auth header, and that 204s return null. Every endpoint in client.rs has a corresponding test.

  Unit tests scattered in command files – 11 tests
  - `common.rs` – parse_app_type validation (accepts known types case-insensitively, rejects unknown)
  - `app/update.rs` – patch serialization
  - `source/upload.rs` – ZIP upload multipart construction

### DIFF
--- a/rust/src/hosting/app/create.rs
+++ b/rust/src/hosting/app/create.rs
@@ -23,8 +23,9 @@ pub(super) fn command() -> RuntimeCommandSpec {
                 "Provision a new hosting application slot. Because no app ID exists \
                  until provisioning completes, this returns an operation ID. \
                  Poll `hosting operation get --operation-id <id>` until \
-                 status is COMPLETED or FAILED. The completed operation result \
-                 includes the created application with its ID.",
+                 status is COMPLETED or FAILED. On COMPLETED, the operation's \
+                 `application` field carries the created app; use `application.id` \
+                 as the --app-id for all subsequent calls.",
             )
             .with_system("hosting")
             .with_tier(Tier::Mutate)
--- a/rust/src/hosting/app/create.rs
+++ b/rust/src/hosting/app/create.rs
@@ -1,0 +1,65 @@
+use cli_engine::{CommandResult, CommandSpec, NextActionParam, RuntimeCommandSpec, Tier};
+use serde_json::json;
+
+use crate::hosting::common::{client_err, make_client, parse_app_type};
+use crate::next_action::next_action;
+use crate::scopes::HOSTING_APPLICATION_CREATE as APP_CREATE;
+
+#[derive(Debug, Clone, clap::Args)]
+struct AppCreateArgs {
+    /// Application type (NODEJS).
+    #[arg(long = "app-type", value_name = "TYPE", value_parser = parse_app_type)]
+    app_type: String,
+
+    /// Human-readable display name (1–200 characters).
+    #[arg(long, value_name = "NAME")]
+    name: String,
+}
+
+pub(super) fn command() -> RuntimeCommandSpec {
+    RuntimeCommandSpec::new_typed_with_context::<AppCreateArgs, _, _, _>(
+        CommandSpec::from_args::<AppCreateArgs>("create", "Create a hosting application")
+            .with_long(
+                "Provision a new hosting application slot. Because no app ID exists \
+                 until provisioning completes, this returns an operation ID. \
+                 Poll `hosting operation get --operation-id <id>` until \
+                 status is COMPLETED or FAILED. The completed operation result \
+                 includes the created application with its ID.",
+            )
+            .with_system("hosting")
+            .with_tier(Tier::Mutate)
+            .mutates(true)
+            .with_scopes(&[APP_CREATE]),
+        |ctx, args: AppCreateArgs| async move {
+            let app_type = args.app_type;
+            let name = args.name;
+            let client = make_client(&ctx, &[APP_CREATE]).await?;
+            let data = client
+                .create_app(&app_type, json!({ "name": name }))
+                .await
+                .map_err(client_err)?;
+
+            let operation_id = data
+                .get("operationId")
+                .and_then(|v| v.as_str())
+                .unwrap_or("")
+                .to_owned();
+
+            let mut poll_action = next_action(
+                "hosting operation get --operation-id <operation-id>",
+                "Poll app provisioning status",
+            )
+            .with_param("operation-id", NextActionParam::required());
+
+            if !operation_id.is_empty() {
+                poll_action = next_action(
+                    "hosting operation get --operation-id <operation-id>",
+                    "Poll app provisioning status",
+                )
+                .with_param("operation-id", NextActionParam::value(operation_id));
+            }
+
+            Ok(CommandResult::new(data).with_next_actions(vec![poll_action]))
+        },
+    )
+}
--- a/rust/src/hosting/app/create.rs
+++ b/rust/src/hosting/app/create.rs
@@ -1,7 +1,7 @@
 use cli_engine::{CommandResult, CommandSpec, NextActionParam, RuntimeCommandSpec, Tier};
 use serde_json::json;
 
-use crate::hosting::common::{client_err, make_client, parse_app_type};
+use crate::hosting::common::{HostingAppOperation, client_err, make_client, parse_app_type};
 use crate::next_action::next_action;
 use crate::scopes::HOSTING_APPLICATION_CREATE as APP_CREATE;
 
@@ -29,7 +29,8 @@ pub(super) fn command() -> RuntimeCommandSpec {
             .with_system("hosting")
             .with_tier(Tier::Mutate)
             .mutates(true)
-            .with_scopes(&[APP_CREATE]),
+            .with_scopes(&[APP_CREATE])
+            .with_output_schema::<HostingAppOperation>(),
         |ctx, args: AppCreateArgs| async move {
             let app_type = args.app_type;
             let name = args.name;
--- a/rust/src/hosting/app/create.rs
+++ b/rust/src/hosting/app/create.rs
@@ -24,7 +24,7 @@ pub(super) fn command() -> RuntimeCommandSpec {
                  until provisioning completes, this returns an operation ID. \
                  Poll `hosting operation get --operation-id <id>` until \
                  status is COMPLETED or FAILED. On COMPLETED, the operation's \
-                 `application` field carries the created app; use `application.id` \
+                 `app` field carries the created app; use `app.id` \
                  as the --app-id for all subsequent calls.",
             )
             .with_system("hosting")
--- a/rust/src/hosting/client.rs
+++ b/rust/src/hosting/client.rs
@@ -1,0 +1,488 @@
+use reqwest::{Client, Method};
+use serde_json::{Value, json};
+
+use crate::application::client::make_http_client;
+
+const BASE_PATH: &str = "/v1/hosting";
+
+#[derive(Debug, thiserror::Error)]
+pub enum ClientError {
+    #[error("HTTP error {status}: {body}")]
+    Http { status: u16, body: String },
+    #[error("network error: {0}")]
+    Network(#[from] reqwest::Error),
+}
+
+impl From<ClientError> for crate::error::GddyError {
+    fn from(value: ClientError) -> Self {
+        match value {
+            ClientError::Http { status, body } => Self::from_http(status, body, "hosting"),
+            ClientError::Network(e) => {
+                Self::network(format!("network error: {e}")).with_system("hosting")
+            }
+        }
+    }
+}
+
+pub struct HostingClient {
+    client: Client,
+    base_url: String,
+    token: String,
+}
+
+impl HostingClient {
+    pub fn new(base_url: impl Into<String>, token: impl Into<String>) -> Self {
+        Self {
+            client: make_http_client(),
+            base_url: base_url.into(),
+            token: token.into(),
+        }
+    }
+
+    fn url(&self, path: &str) -> String {
+        format!("{}{BASE_PATH}{path}", self.base_url)
+    }
+
+    fn new_request_id() -> String {
+        uuid::Uuid::new_v4().to_string()
+    }
+
+    async fn send_json(
+        &self,
+        method: Method,
+        path: &str,
+        query: &[(&str, String)],
+        body: Option<Value>,
+    ) -> Result<Value, ClientError> {
+        let mut req = self
+            .client
+            .request(method, self.url(path))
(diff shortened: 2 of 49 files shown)


rust/src/hosting/app/create.rs
```diff
@@ -0,0 +1,67 @@
...
+use serde_json::json;
+
+use crate::hosting::common::{HostingAppOperation, client_err, make_client, parse_app_type};
+use crate::next_action::next_action;
+use crate::scopes::HOSTING_APPLICATION_CREATE as APP_CREATE;
+
+#[derive(Debug, Clone, clap::Args)]
+struct AppCreateArgs {
+    /// Application type (NODEJS).
+    #[arg(long = "app-type", value_name = "TYPE", value_parser = parse_app_type)]
```
jpage-godaddy: Should this be an enum type instead? I'm also not sure why `NODEJS` is in parentheses; is that implying that only that application type is allowed or is it an example?

rust/src/hosting/app/create.rs
```diff
@@ -0,0 +1,67 @@
+use cli_engine::{CommandResult, CommandSpec, NextActionParam, RuntimeCommandSpec, Tier};
+use serde_json::json;
+
+use crate::hosting::common::{HostingAppOperation, client_err, make_client, parse_app_type};
+use crate::next_action::next_action;
+use crate::scopes::HOSTING_APPLICATION_CREATE as APP_CREATE;
+
+#[derive(Debug, Clone, clap::Args)]
+struct AppCreateArgs {
+    /// Application type (NODEJS).
```
jpage-godaddy: This parenthetical isn't clear. Is `NODEJS` the only option? Could this argument maybe be an enum instead?

rust/src/hosting/guides/hosting.md
jpage-godaddy, line 1: I'd recommend removing these manual line breaks; the markdown renderer for human views applies terminal-width-based word wrapping; manual line breaks interfere with that. So basically, each paragraph should be one line of code.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
+---
+summary: Deploy a Node.js app to GoDaddy hosting — provision, upload, preview, then publish
+---
+
+# `gddy hosting` — deploying a Node.js application
+
+`gddy hosting` manages the full lifecycle of a hosted Node.js application: create an app slot,
+upload source code, test on a staging URL, attach a billing plan, and publish to production.
+Do not use `gddy hosting nodejs`; that is the previous API.
```
jpage-godaddy: Can we just delete those commands? This never left beta, so there's no need to keep it around.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
...
+
+Choose one approach.
+
+**From a local zip archive:**
+
+```sh
+gddy hosting source upload --app-id <app-id> --file ./app.zip
+```
+
+**From a GitHub branch** (connect GitHub at godaddy.com first):
```
jpage-godaddy: What's meant by "connect GitHub at godaddy.com?"

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
...
+```sh
+gddy hosting subscription get --app-id <app-id>
+```
+
+If none is attached yet, list available plans. Only plans with `availableSlots > 0` can take
+a new app. An app attaches to **one** plan. Before `hosting subscription attach`, ask the
+customer to confirm the plan even when only one has slots (show label, tier, slots left).
+If several have slots, let them pick. Do not attach the first row, and do not attach the
+only open plan without confirmation. The list response's `next_actions` includes those
+subscription IDs as an `enum` on `--subscription-id`.
```
jpage-godaddy: These instructions should be written for the user, not addressed to an agent. Sure, agents can read this, but it's meant to be readable by humans as well.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
...
+
+## 4. Attach a subscription (first deploy only)
+
+A hosting plan subscription is required before publishing. Check whether one is already attached:
+
+```sh
+gddy hosting subscription get --app-id <app-id>
+```
+
+If none is attached yet, list available plans. Only plans with `availableSlots > 0` can take
```
jpage-godaddy: What is meant by "list available plans?" Is "plan" a synonym for "subscription?" We should use consistent terminology.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
+---
+summary: Deploy a Node.js app to GoDaddy hosting — provision, upload, preview, then publish
+---
+
+# `gddy hosting` — deploying a Node.js application
+
+`gddy hosting` manages the full lifecycle of a hosted Node.js application: create an app slot,
+upload source code, test on a staging URL, attach a billing plan, and publish to production.
+Do not use `gddy hosting nodejs`; that is the previous API.
+
```
jpage-godaddy, line 8: I would suggest a "Concepts" section before the how-to guides. This could explain subscriptions, slots, apps, and environments so that the how-to guides later on can be streamlined.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
...
+## 6. Custom domains
+
+Attach after the app is published:
+
+```sh
+gddy hosting domain attach --app-id <app-id> --hostname www.example.com
+gddy hosting domain get --app-id <app-id> --domain-id <domain-id>
+```
+
+`domainType` is `PREFIX` (platform subdomain) or `CUSTOM` (customer hostname). PREFIX needs no
```
jpage-godaddy: What is `domainType`? The prior examples make no mention of it. It's not clear what this paragraph is talking about.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,139 @@
...
+summary: Deploy a Node.js app to GoDaddy hosting — provision, upload, preview, then publish
+---
+
+# `gddy hosting` — deploying a Node.js application
+
+`gddy hosting` manages the full lifecycle of a hosted Node.js application: create an app slot,
+upload source code, test on a staging URL, attach a billing plan, and publish to production.
+Do not use `gddy hosting nodejs`; that is the previous API.
+
+Every application has two environments: **PREVIEW** (staging) and **PUBLISH** (production). Source
```
jpage-godaddy: It seems like they're called "environments" in some places but "variants" in another. The guide should be clear to establish terminology, like that you specify environments with `--variant` flags.

rust/src/hosting/client.rs
jpage-godaddy, line 1: I'd recommend using progenitor like how we did for the domain API clients so that we don't have to traverse untyped JSON structures but can instead generate strongly-typed data structures; this could help prevent type mismatch bugs.

jpage-godaddy (commented):
Code is well structured. My biggest concerns are the guide and help text, so a pass through that would be great. Non-blocking, but I do plan on eventually making it mandatory that we generate clients from specs rather than using manual JSON operations, so if you have the time to explore that before release that'd be excellent.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: Agreed. Will update as suggested.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: `PREFIX` is talking about a custom edit to the platform-assigned hostname.

A typical platform-assigned URL would be something like https://{app_id}.c4.test-airoapp.ai. A `PREFIX` domainType would be the internal hostname with the {app_id} part customised.

I will update this for clarity.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: Yeah, I like this idea. I will add it.

rust/src/hosting/app/create.rs
jhenderson1-godaddy: I will change this to an enum since it will only be a few weeks/months before other options become available.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: On reflection this probably doesn't belong in this part of the guide.

Since the hosting CLI won't support directly initiating connections to GitHub at launch, I will move it down into "other commands" since this command will only be relevant for existing apps that were already connected to GitHub through the hosting web UI.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: Updated to improve consistency.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: Agreed. I've removed them now that they're no longer needed.

rust/src/hosting/client.rs
jhenderson1-godaddy: I've taken a look at this and it's definitely a much better approach in terms of consistency and reliability.

Given this would mean a number of changes which would pair well with your general point about generating clients from the spec and keeping them in sync, I've filed a ticket to address this.

Happy to pick it up if you think it's a must before launch, but if not we can keep it as a fast follow.

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,124 @@
...
+
+`gddy hosting` manages the full lifecycle of a hosted Node.js application: create an app, upload source code, test on a staging URL, attach to a hosting plan, and publish to production.
+
+## Concepts
+
+**App** — the hosted Node.js application. Create it first. Later commands take `--app-id`.
+
+**Environments** — each app has PREVIEW (staging) and PUBLISH (production). The CLI calls this `variant`. Uploads land on PREVIEW. `deployment publish` promotes that build to PUBLISH.
+
+**Subscription** — the API name for a hosting plan you already bought (resources and billing). Attach the app to one subscription before the first publish. One app, one subscription.
```
jpage-godaddy: I find this "One app, one subscription." line a bit confusing. Conceptually, it sounds like a subscription has a number of slots, and an app takes up one slot, so that implies to me that you can have one subscription with multiple apps rather than this being a 1:1 relationship (otherwise, why wouldn't we combine these two entities?).

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,124 @@
...
+
+**Slots** — how many more apps that subscription can take (`availableSlots`). Not a separate object. Attach only to a subscription with a free slot.
```
jpage-godaddy:
```suggestion
**Slot** — available space within a subscription for an attached app. Each app attached to a subscription takes up one slot.
```

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,124 @@
...
+
+Poll until COMPLETED, then note the `app.id` in the response:
+
+```sh
+gddy hosting operation get --operation-id <operation-id>
+```
+
+## 2. Upload source code
+
+You can upload again later to the same app. Each upload replaces PREVIEW.
```
jpage-godaddy:
```suggestion
Deploys code to the app's PREVIEW environment. You can run this again to replace what was uploaded before.
```

rust/src/hosting/guides/hosting.md
```diff
@@ -0,0 +1,124 @@
...
+
+## 4. Attach a subscription (first deploy only)
+
+Check whether the app is already on a subscription:
+
+```sh
+gddy hosting subscription get --app-id <app-id>
+```
+
+If not, list subscriptions and attach the app to one with open slots (`availableSlots > 0`). Choose which subscription first, even if only one has a slot.
```
jpage-godaddy: This is worded a bit strange; it seems like it's criticizing the fact that you have to  pick one even if there's only one valid choice.


```suggestion
If not, list subscriptions and attach the app to one with open slots (`availableSlots > 0`).
```

jpage-godaddy (commented):
Just another round of optional comments on the guide; rest looks good.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: It was originally intended as a subtle hint (mostly addressed to agents) to confirm with the customer before attaching the app blindly to any available slot. But it does seem a little redundant and adds friction.

rust/src/hosting/guides/hosting.md
jhenderson1-godaddy: Yeah, your mental model here is right. This line is muddying the water a little. I'll edit it for clarity.

jpage-godaddy (approved):

PR #249, opened 2026-09-04
feat(db): add experimental `db tunnel` (MySQL over WebSocket)

mcolakovic-godaddy:
## Summary

Adds `gddy db tunnel`, an experimental streaming command that opens a local TCP port and bridges raw MySQL bytes over a single WebSocket per connection to an application's agent, which in turn dials the application's own database. This lets a developer point any local MySQL client (`mysql`, TablePlus, DataGrip, an ORM) at a hosted app's database without exposing that database publicly.

Full design and rationale: [`docs/proposals/db-tunnel.md`](docs/proposals/db-tunnel.md).

## How it works

- **Byte pump, not a proxy.** The tunnel never parses the MySQL protocol and never injects credentials. MySQL authentication and TLS are negotiated end-to-end between the local client and the database, so the CLI and the agent see only opaque bytes.
- **One WebSocket per TCP connection.** Each accepted local connection opens its own WebSocket; frames are binary and carry raw protocol bytes in both directions.
- **Token mint.** The command authenticates the one-time token mint with the CLI's own OAuth credential, then receives a short-lived, app-scoped token and the agent URL from the hosting API and presents that token to the agent. No long-lived secrets are stored.
- **Stage-gated.** The `db` group is registered behind the experimental stage and hidden at the GA default, so it does not appear in normal help output.

## Security

- No credential injection or interception — DB auth and TLS are end-to-end between client and database.
- A short-lived, app-scoped mint token is obtained per invocation via the existing OAuth credential, presented to the agent, and never persisted.
- The local listener binds to loopback.

## Testing

- `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (757 tests), `cargo check --locked`, and the module-size check all pass.
- Unit tests cover WebSocket URL construction (ws/wss, with and without an explicit port) and the token-mint request/response contract.

## Notes

- This is the client half of the feature. The server-side token mint and the agent's WebSocket handler are delivered separately and are not required for this change to build or ship — the command is inert until those are available and the experimental stage is enabled.

### DIFF
--- a/docs/proposals/db-tunnel.md
+++ b/docs/proposals/db-tunnel.md
@@ -1,0 +1,292 @@
+# Proposal: `gddy db tunnel` — MySQL access over a WebSocket bridge
+
+Status: draft; implemented behind an experimental feature flag, seeking review
+alongside the server-side changes it depends on.
+
+## Motivation
+
+Developers and support staff sometimes need to point an ordinary MySQL client
+(`mysql`, DBeaver, an app's ORM) at the database behind a GoDaddy-hosted
+application — to inspect data, run a migration, or debug an issue. There is no
+first-class way to do that from `gddy` today.
+
+`gddy db tunnel` opens a local TCP port and, for each client connection, bridges
+raw MySQL bytes over a single WebSocket to the application's per-app **agent**,
+which dials the app's own database. The tunnel is a byte pump: it never parses
+the MySQL protocol and never injects or inspects credentials, so MySQL
+authentication and TLS are negotiated **end-to-end** between the client and the
+real database, exactly as they would be for a direct connection.
+
+## How it works
+
+There are two planes:
+
+- **Control plane** — a single, out-of-band HTTPS call the CLI makes to the
+  GoDaddy hosting API to obtain a short-lived, app-scoped token and the URL of
+  the app's agent. This happens once, when the command starts.
+- **Data plane** — the byte relay. For each local TCP connection the CLI opens
+  one WebSocket to the agent and pumps raw MySQL bytes in both directions. No
+  database traffic ever touches the hosting API.
+
+```mermaid
+flowchart LR
+    client["MySQL client\n(mysql, DBeaver, app)"]
+    subgraph local["Developer machine"]
+        cli["gddy db tunnel\n(local TCP listener)"]
+    end
+    api["GoDaddy hosting API\n(token mint)"]
+    agent["App agent\n(WebSocket endpoint)"]
+    db[("App MySQL\nhost:port")]
+
+    cli -.->|"① mint once: OAuth Bearer → {agent URL, token}"| api
+    client -->|"TCP :3306"| cli
+    cli -->|"② one WSS per TCP conn\nbinary frames = raw MySQL"| agent
+    agent -->|"connect(host, port)\nno credential injection"| db
+```
+
+Key properties, all enforced in code:
+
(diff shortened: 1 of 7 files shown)


docs/proposals/db-tunnel.md
```diff
@@ -0,0 +1,292 @@
...
+    client["MySQL client\n(mysql, DBeaver, app)"]
+    subgraph local["Developer machine"]
+        cli["gddy db tunnel\n(local TCP listener)"]
+    end
+    api["GoDaddy hosting API\n(token mint)"]
+    agent["App agent\n(WebSocket endpoint)"]
+    db[("App MySQL\nhost:port")]
+
+    cli -.->|"① mint once: OAuth Bearer → {agent URL, token}"| api
+    client -->|"TCP :3306"| cli
```
jpage-godaddy: Are we doing TLS for this connection? I worry about data interception.

jpage-godaddy (approved):
This looks great.

rust/src/db/tunnel.rs
```diff
@@ -0,0 +1,604 @@
...
+            )));
+        }
+    };
+    let host = parsed
+        .host_str()
+        .ok_or_else(|| GddyError::network(format!("agent URL '{agent_url}' has no host")))?;
+    let authority = match parsed.port() {
+        Some(port) => format!("{host}:{port}"),
+        None => host.to_owned(),
+    };
```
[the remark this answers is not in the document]
mcolakovic-godaddy: Confirmed: for a bracketed IPv6 literal authority, `host_str()` returns the address without brackets, so the reassembled `wss://` authority would be malformed.

The tunnel resolves its endpoint from a service-minted agent URL, which is always a DNS hostname; a bracketed IPv6 literal only arises when `--agent-url` is pointed by hand at a raw address. The minted-hostname path that ships is therefore unaffected, so this is a correctness gap limited to the manual `--agent-url` override rather than a defect reachable on the normal flow.

jpage-godaddy (approved):

rust/src/scopes.rs
```diff
@@ -130,6 +130,11 @@ declare_scopes! {
     HOSTING_SUBSCRIPTION_READ => "hosting.subscription:read",
     /// Attach an an application to a hosting plan (`hosting subscription attach`).
     HOSTING_SUBSCRIPTION_WRITE => "hosting.subscription:write",
+    /// Open a raw MySQL-over-WebSocket tunnel to a hosting app's database
+    /// (`db tunnel`). Required *in addition to* [`HOSTING_DEPLOYMENT_EXECUTE`],
+    /// so that authority to publish a deployment does not by itself grant raw
+    /// database read/write — opening a tunnel needs this dedicated grant as well.
+    HOSTING_DATABASE_TUNNEL => "hosting.database:tunnel",
```
jpage-godaddy, line 137: Not seeing this scope; blocking merge until we know this is a real scope and have it in the allow list for the CLI.

jpage-godaddy (changes requested):

rust/src/scopes.rs
daleksic-godaddy: @jpage-godaddy We added `hosting.database.tunnel:execute` and configured it https://authz.int.gdcorp.tools/. I believe it has to be linked to cli OAuth app as well.

jpage-godaddy: Great. I've added that scope to the CLI.

jpage-godaddy (approved):

PR #253, opened 2026-09-09
feat(shopping): add Order Management API commands

sswaminathan-godaddy:
## Objective

Expose a beta-gated Shopping customer journey through `gddy`: discover products, create and review a checkout session, accept the checkout's required agreements, place an order, and retrieve the completed order.

## Customer workflow

- `shopping catalog categories`, `search`, `lookup`, and `get`
- `shopping checkout create`, `get`, `update`, and `complete`
- `shopping order get`

Shopping is registered behind `Stage::Beta` at the top-level module, so it remains hidden at the default GA stage. Promoting it later is a one-line removal of that module feature flag.

### Catalog and checkout experience

- API-derived catalog categories, excluding domains, cached for six hours with `catalog categories --refresh`.
- Customer-facing product and purchase-option rendering, checkout-session terminology, structured checkout inputs, buyer details, and contextual next actions.
- Checkout update preserves the current writable checkout state for the API's full-replacement PUT semantics, then retrieves the canonical persisted session before rendering it.
- Payment selection validation rejects unavailable saved payment methods and detects a backend-side silent payment-method substitution after update.
- Checkout create, get, and update render the selected payment method, eligible saved methods (first five by default; all with `--show-all-payment-instruments`), buyer details, totals, links, and required agreements.

### Agreement consent

- Completion retrieves the current checkout before ordering.
- Required agreements are disclosed from the checkout's authoritative `required_agreements` response field, including title, key, URL, and required state.
- `checkout complete` requires `--agree`; without it, the CLI names each required agreement and provides its URL.
- With `--agree`, the CLI sends every distinct required agreement key in `consent.agreement_types`, with a current UTC RFC 3339 `consent.agreed_at` timestamp.
- No consent object is sent for a checkout with no required agreements.
- Completion continues to use the selected saved payment method by default; `--payment-instrument` is optional when a method is already selected.

### Generated client and request safety

- Refreshed the normalized Shopping OpenAPI input and code-generation normalization so the typed client preserves completion consent and checkout required-agreement fields.
- Shopping mutations use generated `Idempotency-Key` HTTP header support for create, update, and complete. The obsolete completion-body idempotency field was removed after Test verification.
- Public Shopping commands do not accept generic request-body/file inputs; completion retains the scoped `--billing-address` input.

### Documentation and tests

- Updated Shopping command help and the Shopping guide for agreement review and consent behavior.
- Added unit/contract coverage for consent serialization, agreement gating, agreement rendering, and header-based mutation requests.

## Test deployment verification

The Order Management MCP Test deployment completed successfully and now returns required agreements.

Using the Test environment with a scoped PAT, verified the complete Hosting customer journey:

1. Discover Web Hosting products.
2. Create and retrieve a checkout session with `universal_terms_and_conditions` returned in `required_agreements`.
3. Confirm completion without `--agree` is rejected locally before purchase.
4. Update buyer details and verify the full canonical checkout response.
5. Complete with `--agree`, which sent consent using the displayed required agreement key.
6. Retrieve the completed order.

The successful Test order was `2211357178`. Full, unredacted commands and command output are retained locally under ignored `stash/shopping-consent-e2e-20260915/README.md`.

## Validation

Executed from `rust/`:

```text
git diff --check
cargo fmt --check
cargo check --workspace
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace
./scripts/check-module-size.sh
cargo audit --ignore RUSTSEC-2026-0194 --ignore RUSTSEC-2026-0195
```

All checks passed. `cargo test --workspace` passed with 874 tests and no failures. Cargo audit found no vulnerabilities; it reports two existing explicitly allowed dependency warnings.


rust/src/shopping/client.rs
```diff
@@ -0,0 +1,203 @@
...
+            .client
+            .request(method, self.url(path))
+            .bearer_auth(&self.token)
+            .header("x-request-id", uuid::Uuid::new_v4().to_string());
+        if let Some(body) = body {
+            request = request.json(&body);
+        }
+        let request = request.build()?;
+        cli_engine::transport::debug_log_reqwest_request(&request);
+        let response = self.client.execute(request).await?;
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in the current branch: Shopping requests no longer call the transport debug helpers, so their request and response payloads are not emitted by `--debug transport`.

rust/src/shopping/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+    Module::new("Shopping", |_ctx| {
+        RuntimeGroupSpec::new(
+            GroupSpec::new("shopping", "Browse catalog products and complete purchases").with_long(
+                "Use the direct Order Management Shopping API integration. Every \
+                 command requests catalog, checkout, and order OAuth scopes together, allowing a \
+                 single consent flow for the catalog → checkout → order lifecycle.\n\
+                 \n\
+                 The direct Katana endpoint must be configured with shopping_url in \
+                 ~/.config/gddy/environments.toml (or SHOPPING_URL). PATs do not work against \
+                 this direct service until front-door token exchange is available.\n\
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Obsolete: the private Shopping URL overrides (`SHOPPING_URL`, `<ENV>_SHOPPING_URL`, and `shopping_url`) were removed after Shopping became available through the selected environment’s standard API front door. The current module resolves `api_url` only.

rust/src/shopping/catalog/search.rs
```diff
@@ -0,0 +1,813 @@
...
+        );
+        Ok(command)
+    } else {
+        let encoded_request = serde_json::to_string(request).map_err(|error| {
+            crate::error::GddyError::unexpected(format!(
+                "failed to encode next-page request: {error}"
+            ))
+            .into_cli_error()
+        })?;
+        Ok(format!("catalog search --body '{encoded_request}'"))
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: generated next actions now retain request data as structured parameters rather than interpolating it into shell-quoted commands. This handles apostrophes and avoids shell-specific command construction.

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,195 @@
...
+            require_selected_payment_instrument(&body)?;
+            let idempotency_key = ensure_completion_idempotency_key(&mut body)?;
+            if ctx.dry_run() {
+                return Ok(CommandResult::new(json!({
+                    "action": "dry-run: would complete checkout",
+                    "id": args.id,
+                    "idempotency_key": idempotency_key,
+                    "body": body,
+                })));
+            }
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: the checkout completion dry-run path returns CommandResult::with_dry_run(), allowing envelope and audit consumers to identify the preview reliably.

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,195 @@
...
+                vec![
+                    next_action(
+                        command_for_env(
+                            &ctx.middleware.env,
+                            format!("order get {order_id} --wait"),
+                        ),
+                        "Read the completed order after it becomes visible",
+                    )
+                    .with_param("order_id", required_value(order_id))
+                    .with_param("idempotency_key", required_value(idempotency_key.clone())),
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: the order follow-up uses the <order-id> placeholder with only the matching order-id parameter. The effective idempotency key remains in completion output rather than being exposed as an unsupported order-get parameter.

rust/src/shopping/checkout/create.rs
```diff
@@ -0,0 +1,153 @@
...
+            } else {
+                input.create_body()?
+            };
+            reject_multiple_payment_instruments(&body)?;
+            if ctx.dry_run() {
+                return Ok(CommandResult::new(serde_json::json!({
+                    "action": "dry-run: would create checkout",
+                    "body": body,
+                })));
+            }
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: the checkout-create dry-run path returns CommandResult::with_dry_run().

rust/src/shopping/checkout/update.rs
```diff
@@ -0,0 +1,125 @@
...
+                .into_cli_error());
+            }
+            reject_multiple_payment_instruments(&body)?;
+            if ctx.dry_run() {
+                return Ok(CommandResult::new(serde_json::json!({
+                    "action": "dry-run: would update checkout",
+                    "id": args.id,
+                    "body": body,
+                })));
+            }
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: the checkout-update dry-run path returns CommandResult::with_dry_run().

rust/src/shopping/client.rs
```diff
@@ -0,0 +1,209 @@
...
+        Self {
+            client: make_http_client(),
+            base_url: base_url.into(),
+            token: token.into(),
+        }
+    }
+
+    fn url(&self, path: &str) -> String {
+        format!("{}{BASE_PATH}{path}", self.base_url)
+    }
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: checkout and order identifiers are encoded through the shared encode_path_segment helper before path construction. A regression test covers /, ?, #, and % in identifiers.

rust/src/shopping/common.rs
```diff
@@ -0,0 +1,645 @@
...
+                return Err(GddyError::not_found(format!(
+                    "order {order_id:?} was not visible after {attempts} attempts over {} seconds",
+                    timeout.as_secs_f32()
+                ))
+                .with_fix(format!(
+                    "Run: gddy {}",
+                    crate::shopping::command_for_env(env, format!("order get {order_id} --wait"))
+                ))
+                .into_cli_error());
+            }
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in f4d7ce7: only an exhausted 404 is mapped to the eventual-visibility not_found error. Final 429, 5xx, and network errors retain their client error mapping.

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[arg(value_name = "CHECKOUT_ID")]
+    id: String,
+
+    /// Saved payment instrument ID to use for this purchase.
+    #[arg(long, value_name = "INSTRUMENT_ID")]
+    payment_instrument: Option<String>,
+
+    /// Stable key for this single intended purchase. A UUID is generated when omitted.
+    #[arg(long, value_name = "KEY")]
+    idempotency_key: Option<String>,
```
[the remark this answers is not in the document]
sswaminathan-godaddy: Resolved in the updated PR description and f4d7ce7: --idempotency-key is optional; gddy generates and returns a UUID when omitted, while a caller-supplied nonblank key is preserved for the same intended purchase.

rust/src/shopping/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+        register_order_get_human_view(ctx);
+        RuntimeGroupSpec::new(
+            GroupSpec::new(
+                "shopping",
+                "Browse GoDaddy products, execute checkout, and view completed orders",
+            )
+            .with_long(
+                "Browse GoDaddy products, create/update/complete checkout, and view completed orders.\n\
+                 \n\
+                 Shopping commands request the required OAuth permissions together so you can \
```
jpage-godaddy: What does this mean? That it requests _all_ permissions no matter what you do? I am not sure users would understand this text.

rust/src/shopping/guides/shopping.md
jpage-godaddy, line 1: I'd suggest soft line breaks, not hard. That's because the markdown renderer takes care of line wrapping based on the terminal width. With explicit hard line breaks, the text is breaking exactly where it is in the markdown file instead, which will not work as well for narrower terminals.

rust/src/shopping/guides/shopping.md
```diff
@@ -0,0 +1,266 @@
+---
+summary: Browse products, place orders, and retrieve purchase details.
+---
+
+# Shopping API
+
+`gddy shopping` integrates with the Shopping API.
```
jpage-godaddy: The CLI is an abstraction over APIs; we should be focused on telling customers what they can do, not which APIs are being called.

jpage-godaddy:
```suggestion
# Shopping for GoDaddy Products

Use `gddy shopping` commands to purchase GoDaddy's products and review your past purchases.
```

rust/src/shopping/guides/shopping.md
```diff
@@ -0,0 +1,266 @@
...
+```bash
+gddy auth login \
+  --scope shopping.catalog:read \
+  --scope shopping.checkout:execute \
+  --scope shopping.order:read
+```
+
+Shopping requests use the selected environment's standard API front door. OAuth requests the
+complete lifecycle scope bundle above in one consent flow. PAT support requires those Shopping
+scopes to be available on the Developer Portal and is tracked separately.
```
jpage-godaddy: Is any of this something a user needs to know? Users aren't going to have any environments besides prod and in edge cases OTE. They're not going to know what "front door" means. This "is tracked separately" doesn't make sense for users either...

I think this entire guide needs a rewrite so it is geared toward users of the CLI. My suggested outline:

1. Introduction, what this guide teaches you (how to buy GoDaddy products)
2. Concepts (what's in a catalog, the cart, payment methods, etc. I am writing this before reviewing the whole thing, so whatever concepts the CLI exposes, not the ones I'm speculating on)
3. How to accomplish specific tasks (command sequence, explain anything non-obvious)

rust/src/shopping/catalog/get.rs
```diff
@@ -0,0 +1,274 @@
...
+            } else {
+                response
+            };
+            Ok(CommandResult::new(output).with_next_actions(actions))
+        },
+    )
+}
+
+fn next_actions(response: &Value, env: &str) -> Vec<cli_engine::NextAction> {
+    let Some(product) = response.get("product") else {
```
jpage-godaddy, line 52: If we have a spec, I highly recommend generating code (using progenitor like we did for domains) so that we have strong types instead of working with typeless JSON. That should avoid potential bugs in JSON traversal and make the code less verbose with all the type conversions.

rust/src/shopping/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Every Shopping operation requests the full lifecycle scope bundle at once.
+/// This deliberately avoids disruptive OAuth consent/step-up during the common
+/// catalog → checkout → order workflow.
+pub(crate) const SHOPPING_SCOPES: &[&str] = &[
+    SHOPPING_CATALOG_READ,
+    SHOPPING_CHECKOUT_EXECUTE,
+    SHOPPING_ORDER_READ,
+];
+
+pub(crate) fn command_for_env(env: &str, command: impl AsRef<str>) -> String {
```
jpage-godaddy: We don't need this. If the user has set their environment with `gddy env set` they don't need to explicitly pass `--env`. No actual users will use environments, so we shouldn't optimize for internal testers' experience.

rust/src/shopping/catalog/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use cli_engine::{GroupSpec, RuntimeGroupSpec};
+
+pub(super) fn group() -> RuntimeGroupSpec {
+    RuntimeGroupSpec::new(
+        GroupSpec::new("catalog", "Browse and resolve Shopping catalog products").with_long(
+            "Search the live Shopping catalog, resolve known product IDs, and retrieve product details. \
+             Every command requests the complete Shopping OAuth scope bundle so a catalog-to-checkout \
+             workflow needs only one consent flow.",
+        ),
```
jpage-godaddy: Does it make sense to capitalize "Shopping?" Just make sure we're not leaking _our system names_ which are not meaningful to our users. "Resolve" doesn't really have an obvious meaning.

```suggestion
        GroupSpec::new("catalog", "Explore GoDaddy's product catalog").with_long(
            "Search the GoDaddy product catalog and retrieve product details.",
        ),
```

rust/src/shopping/catalog/search.rs
```diff
@@ -0,0 +1,878 @@
...
+    "messages": "[]object", optional;
+});
+
+#[derive(Debug, Clone, clap::Args)]
+struct Args {
+    /// Text to search for. Omit to browse the catalog.
+    #[arg(long, value_name = "TEXT")]
+    query: Option<String>,
+
+    /// Product category to include. Repeat to include multiple categories.
```
jpage-godaddy: How do users know what the categories are?

rust/src/shopping/catalog/search.rs
```diff
@@ -0,0 +1,878 @@
...
+
+    /// Maximum number of products to return (1-100).
+    #[arg(long, value_name = "N", value_parser = clap::value_parser!(u8).range(1..=100))]
+    limit: Option<u8>,
+
+    /// Preferred ISO 4217 currency for returned catalog prices (for example, USD or GBP).
+    #[arg(long, value_name = "CODE", value_parser = currency_code)]
+    currency: Option<String>,
+
+    /// Buyer country used for catalog eligibility and pricing.
```
jpage-godaddy: 1. (really for the API) is this an honor code system? Why do our users supply a country code?
2. Needs examples (not everyone knows what an "ISO" country code is, and plus there are multiple ISO standards for this).
3. The user is the buyer; the help text should be oriented toward the user.

rust/src/shopping/catalog/search.rs
```diff
@@ -0,0 +1,878 @@
...
+            "List Price",
+            "Term",
+            "Availability",
+        ],
+        &rows,
+        &[false, false, false, true, true, false, false],
+    )
+}
+
+fn render_table(headers: &[&str], rows: &[Vec<String>], right_aligned: &[bool]) -> String {
```
jpage-godaddy: Why are we manually rendering a table instead of using the built-in CLI engine functionality for this?

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Debug, Clone, clap::Args)]
+struct Args {
+    /// Checkout session ID.
+    #[arg(value_name = "CHECKOUT_ID")]
+    id: String,
+
+    /// Saved payment instrument ID to use for this purchase.
+    #[arg(long, value_name = "INSTRUMENT_ID")]
+    payment_instrument: Option<String>,
```
jpage-godaddy, line 24: How do users get a payment instrument ID?

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,244 @@
...
+        .with_fix(format!(
+            "Completion may have reached Shopping. Do not retry automatically. Reuse idempotency_key \
+             {idempotency_key:?} only for the same intended purchase after confirming its outcome."
+        ))
+        .into_cli_error()
+}
+
+pub(super) fn command() -> RuntimeCommandSpec {
+    RuntimeCommandSpec::new_typed_with_context::<Args, _, _, _>(
+        CommandSpec::from_args::<Args>("complete", "Complete a Shopping checkout and place an order")
```
jpage-godaddy:
```suggestion
        CommandSpec::from_args::<Args>("complete", "Place an order with the contents of your cart")
```

rust/src/shopping/checkout/complete.rs
```diff
@@ -0,0 +1,244 @@
...
+            .with_long(
+                "Places a real order. Use --payment-instrument for one saved payment instrument, \
+                 or --body/--file for advanced payment or billing-address fields. Use --idempotency-key \
+                 to control retries, or omit it to let gddy generate and return one. The CLI never retries \
+                 completion automatically. Read the resulting order with `shopping order get <order-id> --wait`.",
```
jpage-godaddy: As opposed to a fake order? The "real" seems odd...

I'm not sure if we need all this noise about idempotency keys or retries up front; instead, if the operation fails, the "fix" text of the error should give the suggested step to do the retry which I assume will contain the idempotency key.

rust/src/shopping/checkout/create.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Path to a JSON checkout-create request. Takes precedence over --body.
+    #[arg(long, value_name = "PATH")]
+    file: Option<String>,
+}
+
+pub(super) fn command() -> RuntimeCommandSpec {
+    RuntimeCommandSpec::new_typed_with_context::<Args, _, _, _>(
+        CommandSpec::from_args::<Args>("create", "Create a Shopping checkout session")
+            .with_long(
+                "Create a checkout with --item and optional buyer/payment flags. Repeat --item for \
```
jpage-godaddy: I'm wondering why we don't use common commerce terms like "cart" or "basket." A "checkout" is an action. When I read "create a checkout," to me that means something more like "complete an order" not "create an empty cart." Also, do we have the concept of a default cart? Ideally, customers can just add products they want to buy to _a_ cart and then check out/purchase the contents of the cart.

rust/src/shopping/checkout/get.rs
```diff
@@ -0,0 +1,365 @@
...
+        "selected_payment": selected_payment(checkout),
+        "available_payment_instruments": available_payment_instruments(checkout, show_all_payment_instruments),
+        "has_more_payment_instruments": !show_all_payment_instruments && available_payment_instrument_count(checkout) > PAYMENT_INSTRUMENT_LIMIT,
+        "next_steps": human_next_steps(actions),
+    })
+}
+
+const PAYMENT_INSTRUMENT_LIMIT: usize = 5;
+
+fn available_payment_instruments(checkout: &Value, show_all: bool) -> Vec<Value> {
```
jpage-godaddy: We should have a separate command for listing payment instruments.

rust/src/shopping/guides/shopping.md
```diff
@@ -0,0 +1,266 @@
...
+  --buyer-phone '+15550100'
+```
+
+Repeat `--item` to create a cart; append `=QUANTITY` to an item, such as
+`--item '<available-variant-id>=2'`. Add `--payment-instrument <saved-payment-instrument-id>`
+to select one stored payment method. It is optional at creation: a ready checkout can expose a
+saved instrument for selection at completion.
+
+A stored payment instrument normally supplies its saved billing address automatically. Use a
+JSON document when you need an address override or other advanced nested fields:
```
jpage-godaddy: I'm not sure a sample JSON document makes sense. If JSON is being used, that would be something an AI agent would probably need to do for the human, and the AI agent should have a schema so it knows the shape of the JSON. Same for all these other advanced options.

I'd suggest either omitting these JSON/file parameters or at least having the guide link agents to where they can get the JSON schema.

rust/src/shopping/client.rs
jpage-godaddy, line 1: Why not use the progenitor pattern here instead of manually traversing JSON and risking mismatch with the spec?

rust/src/shopping/money.rs
```diff
@@ -0,0 +1,107 @@
+use serde_json::Value;
+
+/// Shopping amounts are ISO-4217 minor units. Their decimal scale derives from the returned
```
jpage-godaddy: Make sure we're not duplicating the currency formatting we already have for domains. If it's the same standard, we're using a library that understands the standard.

rust/src/next_action.rs
```diff
@@ -25,14 +26,55 @@ pub(crate) fn required_value(value: impl Into<String>) -> NextActionParam {
     }
 }
 
+/// Mirrors cli-engine placeholder substitution for custom human views. The
+/// structured template and parameters remain in the output envelope.
```
jpage-godaddy: Claude... why do you keep "mirroring" functionality (duplicating) instead of reusing what we already have? We already have a standard output envelope to include these actions in a standard way. If we really need to duplicate code, we need a better explanation on why that is necessary.

rust/src/shopping/mod.rs
sswaminathan-godaddy: Makes sense... Probably unnecessary.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Will do.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: will update

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: SOunds good. Will take another stab at it.

rust/src/shopping/catalog/get.rs
sswaminathan-godaddy: ok, will look into this

rust/src/shopping/mod.rs
sswaminathan-godaddy: ok

rust/src/shopping/catalog/mod.rs
sswaminathan-godaddy: Agreed. I blame Claude.

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: Right now, UCP doesn't have a definitive way and we haven't implemented any extensions to discover categories available. We can potentially hard code some categories in the CLI until we have such an API

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: Claude's choice - will have it fix it...

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: 1. yeah, even FOS behaves like this. Users don't have to specify country
2. Will fix
3. Agreed - will address.

rust/src/next_action.rs
sswaminathan-godaddy: Agreed! will have it take a closer look.

rust/src/shopping/money.rs
sswaminathan-godaddy: Will address

rust/src/shopping/client.rs
sswaminathan-godaddy: Sounds good (I personally had no idea there's such a thing - just went with Claude's impl as long as the testing achieved the outcome)

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: They can currently only get it when creating a checkout - that's kind of how UCP does it... I'll get claude to document the flow better

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: Agreed!

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: lol :) Agreed. Will address.

rust/src/shopping/checkout/create.rs
sswaminathan-godaddy: Makes sense. Will rework this.

rust/src/shopping/checkout/get.rs
sswaminathan-godaddy: yeah... ok, I can put one in

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: ok, maybe just stick to simple inputs all the way through then... at least for now... will address

rust/src/shopping/checkout/get.rs
sswaminathan-godaddy: I agree, but it's a bit complicated.. Applicable payment methods are really specific to the checkout context. For instance, if the currency were to change, it could potentially change the available payment options - this is how FOS works as well (and perhaps why UCP doesn't have a separate way to get available payments).

If anything, let's defer explicit commands for this for first release. I will try to make the guide better to help users understand the flow easily.

sswaminathan-godaddy: If we want, I can expose a command that lists available payments in a checkout session.

rust/src/shopping/mod.rs
sswaminathan-godaddy: Removed the customer-facing permission and scope explanation; required scopes remain internal command metadata.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Rewrote guide prose using soft wraps; explicit continuations remain only in shell examples.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Reworked the guide around customer outcomes: discover products, manage a cart, place an order, and review purchases.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Rewrote the guide as a customer-focused catalog-to-cart-to-order journey and removed internal environment/front-door material.

rust/src/shopping/catalog/get.rs
sswaminathan-godaddy: Added a generated shopping-client crate backed by the authoritative Shopping OpenAPI contract and Progenitor builders.

rust/src/shopping/mod.rs
sswaminathan-godaddy: Removed environment injection from Shopping follow-up commands; the selected CLI environment is used implicitly.

rust/src/shopping/catalog/mod.rs
sswaminathan-godaddy: Applied the suggested customer-facing catalog wording.

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: The CLI now validates a maintained non-domain category vocabulary and documents supported values with examples.

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: Removed --country. Currency help now uses concrete USD/GBP examples and catalog search is customer-oriented.

rust/src/shopping/catalog/search.rs
sswaminathan-godaddy: Replaced manual search-table formatting with cli-engine HumanViewDef/TableColumn rendering and structured rows.

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: Cart create/get output and the guide now expose contextual saved payment methods and their IDs before completion.

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: Applied the requested command description: Place an order with the contents of your cart.

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: Removed real-order and idempotency/retry guidance from customer-facing help, guides, normal output, and dry-run output.

rust/src/shopping/checkout/create.rs
sswaminathan-godaddy: Kept the stable shopping checkout command path but use cart, payment method, and order consistently in customer-facing help, guide text, and output.

rust/src/shopping/checkout/get.rs
sswaminathan-godaddy: Deferred a standalone payment-method command for this release: eligible methods are cart-specific. Cart create/get exposes the contextual choices and IDs.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Removed raw JSON walkthroughs from the guide; the main journey now uses only simple command inputs. Advanced inputs remain available for exceptional use.

rust/src/shopping/client.rs
sswaminathan-godaddy: Migrated Shopping transport and request modeling to a generated Progenitor client; dynamic JSON remains only at the CLI projection/compatibility boundary.

rust/src/shopping/money.rs
sswaminathan-godaddy: Consolidated ISO-4217 minor-unit formatting on the existing Domains implementation; Shopping retains only its response-shape extraction.

rust/src/next_action.rs
sswaminathan-godaddy: Removed project-local next-action display/substitution logic. cli-engine now renders standard structured next-action footers for custom views.

rust/src/shopping/guides/shopping.md
sswaminathan-godaddy: Update: the later checkout-flow refinement removed the remaining raw `--body`/`--file` inputs from Shopping commands. The only JSON input retained is the constrained `--billing-address` option on completion; the normal flow uses command flags.

rust/src/shopping/checkout/create.rs
sswaminathan-godaddy: Update: the final wording uses `checkout session` consistently for the resource named by the command/API, while explaining in the guide that a checkout session is your cart. It no longer uses “cart” as the primary term.

rust/src/shopping/checkout/complete.rs
sswaminathan-godaddy: Update: the final command description is `Place an order from a checkout session`; the longer help explains the selected payment method and required `--agree` acknowledgement.

rust/src/shopping/checkout/complete.rs
jpage-godaddy: I see. I think it's a rough edge on the experience, but shouldn't block releasing this.

jpage-godaddy (approved):

PR #257, opened 2026-09-12
fix(dns): pair retained values before position in the set plan

AshutoshMore142k4:
## Summary
- `plan_set` paired existing records to `--data` values purely by position, so a value that was *staying* in the set could get paired with a different record and re-created. v3 keys record identity on (name, type, data), so that create fails with `DUPLICATE_RECORD` — and `apply_replace` deliberately keeps the old record when its create fails, while the surplus `Delete` later in the plan still runs unconditionally.
- Net effect on a narrowing set: given `www A {1.2.3.4, 5.6.7.8}`, running

  ```
  gddy dns set example.com --type A --name www --data 5.6.7.8
  ```

  planned `Replace{r1 → 5.6.7.8}`, `Delete{r2}`. The replace's create collided with r2's still-live `5.6.7.8`, so r1 was kept; the delete then removed r2. The zone ended up holding `1.2.3.4` — the value the user asked to drop — with `5.6.7.8` gone. The command does exit non-zero, but the zone is already inverted by then.
- A pure reorder (`--data 5.6.7.8 --data 1.2.3.4` against the same zone) is the same root cause with a milder symptom: both replaces fail as duplicates and the command errors having changed nothing.
- Fix: pair each desired value with the record that already holds it *first*, then pair whatever is left by position exactly as before. A value that stays in the set is therefore never re-created, so the collision can't arise. `plan_set` now takes `(record_id, current value)` so it can see content; ordering of the emitted plan (replaces → deletes → creates) is unchanged.

Ordering deletes before creates would also dodge the collision, but it opens a window where the name resolves to nothing — so this keeps the existing create-then-delete shape.

## Test plan
- [x] `cargo check --workspace`
- [x] `cargo fmt --check`
- [x] `cargo test --workspace` — 780 passed
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean for this change. (Unrelated pre-existing note: on a Windows host it errors on an unused `use super::*` in `extension/security/file_discovery.rs`, whose only two tests are `#[cfg(unix)]`, so that test module compiles empty there. CI on ubuntu is unaffected.)
- [x] Four new `plan_set` cases: the narrowing case above, a pure reorder, a mixed keep-one/replace-one, and duplicate `--data` values pairing to distinct records
- [x] Existing `plan_set` and `dry_run_set_preview` tests updated for the new parameter shape and still assert the same plans

### DIFF
--- a/rust/src/dns/set/mod.rs
+++ b/rust/src/dns/set/mod.rs
@@ -8,8 +8,8 @@ use crate::scopes::DOMAINS_DNS_UPDATE;
 
 use super::records::verify_with_list_action;
 use super::records::{
-    RecordOptions, RecordWriteArgs, fetch_records, validate_caa_fields, validate_svcb_fields,
-    validate_tlsa_fields,
+    RecordOptions, RecordWriteArgs, fetch_records, record_value, validate_caa_fields,
+    validate_svcb_fields, validate_tlsa_fields,
 };
 
 mod outcome;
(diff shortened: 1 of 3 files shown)


jpage-godaddy:
Hi there. Thanks for your contribution (and the one before which I merged in). Can you address these Copilot findings please?

AshutoshMore142k4:
Thanks @jpage-godaddy Jacob, will do. I’ll update the retained-record matching to use the full record identity, in line with `same_content`, instead of just `record_value`.

I’ll also add a couple of tests for record types with additional identity fields, like TLSA/CAA, and push the changes in a follow-up commit. I’ll request another review once it’s ready.

PR #263, opened 2026-09-16
fix(domain): allow PAT auth for domain purchase

fgorodishter-godaddy:
## Problem

`gddy domain purchase` fails for every PAT-authenticated caller, before any HTTP request is made:

```
the OAuth token's subject ("") is not a customer identity;
`domain purchase` needs a customer-scoped token
```

`consent_principal` requires the credential's subject to be `customer:<uuid>`, but `pat_credential` (`src/auth.rs`) leaves `sub` empty by design — a PAT is opaque to the CLI, and its identity is resolved by the gateway. No scope grant can work around this: the check never looks at scopes, and `Credential::scopes` is likewise empty for PATs.

## Why the check was safe to relax

The extracted id was already discarded by the only caller (`let _customer_id = ...`). The register endpoint derives `agreedBy` from the request's auth context, so this was purely a local fail-fast guard — and for PATs it guarded against nothing while blocking a request the API accepts.

Confirmed against `test` by calling `POST /v3/domains/registrations` directly with a PAT: it returned `202` and the response carried a server-derived `agreedBy.principal`, i.e. the gateway resolved the customer the CLI claimed it couldn't find.

## Change

- Exempt PAT credentials from the subject check, keyed on `cred.provider`.
- Rename `consent_principal` → `validate_consent_identity`, returning `Result<()>`, since it no longer produces a principal.

The exemption keys on the provider rather than on an empty `sub`, so an OAuth token with no subject is still rejected. A test pins that down.

## Testing

- `cargo check --workspace`, `cargo clippy --workspace -- -D warnings`, `cargo fmt --check`, `cargo test --workspace` — all clean.
- New test `consent_identity_exempts_pat_credentials_with_no_subject` covers the PAT case and the OAuth-with-empty-subject counterpart.
- End-to-end with a PAT against `test`: registered `e2etitanapi2.com` via `gddy domain quote` → `gddy domain purchase --agree --confirm`, returning `status: COMPLETED`. Before this change the same command failed on the subject check.
- Zero-cost regression probe: `domain purchase --quote-token <bogus>` now reaches the quote-cache lookup ("no cached quote for that token") instead of the identity error, confirming the gate no longer short-circuits.

Note that `--dry-run` would not have caught the original bug — it is evaluated after the credential check.


jpage-godaddy:
More comprehensive removal here:
https://github.com/godaddy/cli/pull/264

PR #266, opened 2026-09-16
feat(import): migrate init from-existing to import command

mguerrero3-godaddy:
After recent `--from-existing` flag addition, `init` was doing two unrelated things: creating a new app (mutating) vs. syncing an existing app's config into godaddy.toml (read-only), gated by a flag. This work splits the second path into its own command, `gddy platform app import <name>`.

- New import command: pulls an app's remote config + latest release's webhook subscriptions into godaddy.toml. Read-only against the API, no app is created, no .env written, safe to re-run.
- init reverts to pure create-a-new-app; --from-existing/--force flags removed.
- New ApplicationImport output schema (separate from ApplicationInit).                                                     

No behavior change to existing paths, `init` unaffected, import preserves the prior `--from-existing` fetch/write/subscription-diff logic.


jpage-godaddy (approved):

smukherjee-godaddy:
Thanks for separating `app import` from `app init`—this addresses the command-design concern from the Slack discussion.

A few issues to address before merging:

1. `init --from-existing` shipped in v0.2.14, so removing it immediately breaks existing scripts. Please keep it temporarily as a deprecated forwarding alias to `app import`.
2. `read_config(&config_path).ok()` suppresses parse, validation, permission, and I/O errors. The command can then overwrite an unreadable manifest and lose local sections. Only `ConfigError::NotFound` should be treated as no existing configuration.
3. Before preserving local actions, extensions, dependencies, settings, or version, verify that the existing manifest belongs to the application being imported. Otherwise, importing app B in app A's directory can silently carry app A's configuration into app B.
4. The help says `app update` can change URL, proxy URL, and scopes, but that command currently supports only label and description. Please correct the guidance or add the missing support.

The new command registration and `Tier::Mutate` classification look correct, and all current GitHub checks pass.

rust/src/platform/app/commands/init.rs
```diff
@@ -45,283 +45,27 @@ struct InitArgs {
...
 
-    /// Fetch an already-registered application's remote config and its
-    /// latest release's webhook subscriptions instead of creating a new
-    /// application. --description, --url, --proxy-url, and --scopes
-    /// override the corresponding fetched value if also provided.
+    /// [DEPRECATED: use `gddy platform app import <name>`] Fetch an
+    /// already-registered application's remote config and its latest
+    /// release's webhook subscriptions into godaddy.toml, instead of
+    /// creating a new application.
     #[arg(
```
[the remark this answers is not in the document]
mguerrero3-godaddy: This is expected due to the flag getting to latest release. Keeping It for some releases to provide information to any current usage.

rust/src/platform/app/commands/init.rs
```diff
@@ -45,283 +45,30 @@ struct InitArgs {
...
+    /// release's webhook subscriptions into godaddy.toml, instead of
+    /// creating a new application. Unlike the old --from-existing, this no
+    /// longer accepts --label/--description/--url/--proxy-url/--scopes
+    /// overrides — like `import`, it strictly mirrors the remote app; use
+    /// `gddy platform app update` to change label/description afterward.
     #[arg(
         long,
         value_name = "NAME",
-        conflicts_with_all = ["accept_agreements", "name", "config"]
+        conflicts_with_all = [
```
smukherjee-godaddy: [P1] The deprecated alias still breaks previously valid invocations. v0.2.14 allowed `--description`, `--url`, `--proxy-url`, and `--scopes` with `--from-existing`; these conflicts make existing scripts fail during argument parsing before they can receive the deprecation warning. If this alias is the compatibility bridge, please keep accepting and forwarding the previously supported overrides until the alias is removed.

mguerrero3-godaddy: I thought since we were proxying to the new command handler this wasn't a problem but was able to verify you're right. Adjusted.

smukherjee-godaddy (approved):
Re-reviewed the latest changes. The command separation, manifest-read protection, cross-application guard, help text, validation mapping, and mutation classification are addressed, and all checks pass. The remaining deprecated-override compatibility concern is non-blocking.

PR #275, opened 2026-09-18
chore(hosting): use WEB_HOSTING as hosting-product flag in next …

jhenderson1-godaddy:
…actions

## Summary

The hosting CLI currently suggests running `hosting subscription list` when a `WH_PLAN_REQUIRED` error is encountered when publishing an app. 

That command requires a `hosting-product` argument, so the command needs to be updated accordingly (this will be type enforced by https://github.com/godaddy/cli/pull/276).


rust/src/hosting/common.rs
```diff
@@ -304,7 +304,7 @@ mod tests {
                 status: 422,
                 body: r#"{"message":"This app must be attached to a Web Hosting plan to publish.","details":[{"issue":"WH_PLAN_REQUIRED"}]}"#.to_owned(),
             },
-            "Run: gddy hosting subscription list",
+            "Run: gddy hosting subscription list --hosting-product=WEB_HOSTING",
```
jpage-godaddy, line 307: Can you also make `--hosting-product` a required flag?

jhenderson1-godaddy: That's already covered [here](https://github.com/godaddy/cli/pull/276), but I can add it in this PR too

jpage-godaddy: Ah, ok. I'll just merge this then.

